PR #8, opened 2020-11-30
swapaxes Numpy API 

kssteven418:
This PR impements numpy `swapaxes(axis1, axis2)` API for NumS `BlockArray` object. swapaxes Numpy API


nums/core/array/blockarray.py
```diff
@@ -160,6 +160,30 @@ def squeeze(self):
...
+        meta_swap["shape"] = tuple(shape)
+        meta_swap["block_shape"] = tuple(block_shape)
+        grid_swap = ArrayGrid.from_meta(meta_swap)
+        rarr_swap = BlockArray(grid_swap, self.system)
+        rarr_src = self.blocks.swapaxes(axis1, axis2).reshape(-1)
+        rarr_tgt = rarr_swap.blocks.reshape(-1)
+        for idx, j in np.ndenumerate(rarr_src):
+            rarr_tgt[idx] = j.copy()
+
+        for grid_entry in rarr_swap.grid.get_entry_iterator():
```
elibol, line 181: Since `swapaxes` is no longer an in-place operation, please update this loop to assign the returned block accordingly.

nums/core/array/blockarray.py
```diff
@@ -160,6 +160,30 @@ def squeeze(self):
             new_block_shape.append(b)
         return self.reshape(new_shape, new_block_shape)
 
+    def swapaxes(self, axis1, axis2):
```
elibol, line 163: This appears correct as-is. Let me propose an alternative:
1. Once you have `grid_swap`, create `rarr_src` without flattening the array. `rarr_src` is now a distinct view on the blocks and doesn't need to be copied again. All that's left is to update the blocks themselves.
2. Iterate over the blocks using `for grid_entry in grid_swap.get_entry_iterator()`. You can update the new view of the blocks `rarr_src` directly using `rarr_src[grid_entry] = rarr_src[grid_entry].swapaxes(axis1, axis2)`.
3. `BlockArray` takes an optional array of blocks as a parameter, so you can directly return `BlockArray(grid_swap, self.system, rarr_src)`.

Of course, rename variables accordingly.

nums/numpy/api.py
```diff
@@ -261,6 +261,45 @@ def logspace(start, stop, num=50, endpoint=True, base=10.0, dtype=None, axis=0):
         ba = ba.astype(dtype)
     return ba
 
+############################################
+# Matrix Ops
+############################################
+
+def tensordot(x1: BlockArray, x2: BlockArray,
```
elibol, line 268: Didn't we already merge this and `matmul`?

elibol (commented):
This looks really good! I left a few comments. Please let me know if they make sense and let me know what you think.

nums/core/array/blockarray.py
kssteven418: I changed the code according to the new `Block.swapaxes` function.
This is not exactly the same as what you proposed, but I think this will work the same as well. 
- I made an empty `rarr_src` numpy array and then initialized each entry with `self.blocks[grid_entry].swapaxex`.

nums/numpy/api.py
kssteven418: I also wonder why it appears in the diff, since I did not do anything here.

elibol: Have you merged the latest `api` branch into your branch? If you have, you may have merged in a way where these changes are marked as more recent. If you've merged and this is still happening, an easy way to correct the issue is to create a new branch from `api` and copy (or cherry pick) your changes over. You can also go back to the point at which `api` was merged into your branch and correct the merge, but this is a more complicated, time consuming, and error prone approach.

elibol:
This is almost done. The most recent changes look correct, but the diff of changes we've already committed is still a problem.

nums/numpy/api.py
kssteven418: That was repetitive code due to the merge error, I guess. I removed one of them.

kssteven418:
I think I forgot to push my last commit:( so now the `swapaxes` function is modified as desired.
Matmul and Shape APIs were repetitive, maybe due to git auto merge issue, so I removed the repetitive codes.

Last remaining step is pylint, which I have trouble running it on my local machine.
Comment that I used is: `pytest --pylint` 
and the output error is:
```
ERROR: usage: py.test [options] [file_or_dir] [file_or_dir] [...]
py.test: error: unrecognized arguments: --pylint
```

Currently, I'm using `pylint 2.6.0` and `pytest 6.1.1`. Are these the only requirement for running pylint, or should I setup something else?

elibol:
> I think I forgot to push my last commit:( so now the `swapaxes` function is modified as desired.
> Matmul and Shape APIs were repetitive, maybe due to git auto merge issue, so I removed the repetitive codes.
> 
> Last remaining step is pylint, which I have trouble running it on my local machine.
> Comment that I used is: `pytest --pylint`
> and the output error is:
> 
> ```
> ERROR: usage: py.test [options] [file_or_dir] [file_or_dir] [...]
> py.test: error: unrecognized arguments: --pylint
> ```
> 
> Currently, I'm using `pylint 2.6.0` and `pytest 6.1.1`. Are these the only requirement for running pylint, or should I setup something else?

You'll also need `pytest-pylint`.

nums/numpy/api.py
elibol: This is still an issue. I'm okay with you creating a new branch from `api`, updating it with changes from this branch, and creating a new PR.

kssteven418:
I opened a new PR, #9

PR #13, opened 2021-01-14
Implemented all function, Updated README

MohamedElgharbawy:
Added double quotes to support pip install on zsh shells (it still works on bash).

### DIFF
--- a/tests/numpy/test_misc.py
+++ b/tests/numpy/test_misc.py
@@ -95,6 +95,32 @@ def test_reshape(nps_app_inst):
                         nps.reshape(ba, shape=(6, 4)))
 
 
+def test_all(nps_app_inst):
+    import nums.numpy as nps
+    assert nps_app_inst is not None
+
+    true_int = nps.array([[1, 2, 3], [1, 2, 3]])
+    false_int = nps.array([[1, 2, 0], [1, 2, 3]])
+    true_bool = nps.array([True, True])
+    false_bool = nps.array([True, False])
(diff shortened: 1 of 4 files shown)


tests/numpy/test_misc.py
```diff
@@ -94,6 +94,15 @@ def test_reshape(nps_app_inst):
     assert nps.allclose(ba.reshape(shape=(6, 4), block_shape=(6, 4)),
                         nps.reshape(ba, shape=(6, 4)))
 
+def test_all(nps_app_inst):
```
elibol, line 97: Can you add a test that evaluates bools and floats directly? Something simple is fine -- just make sure you cover both cases, where `all` should evaluate to true and false.

Also, while the test for int is fine (you can do bool the same way), for the float test, test the output against what numpy does. I'm not sure whether it evaluates a float `0.0` as true or false.

Finally, please add an assertion for the dtype of all tests (e.g. `assert nps.all(b).dtype is np.bool`).

elibol (commented):
This looks perfect. I left a few comments for the test. Great job!

tests/numpy/test_misc.py
MohamedElgharbawy: Added all suggestion.s

elibol: Almost there. Could you add type checks for the bool and int types as you have with int (e.g. `assert nps.all(true_int).dtype is np.bool`)?

MohamedElgharbawy: Done.

PR #14, opened 2021-01-19
Implemented alltrue, any, array_equal

MohamedElgharbawy:
Also removed implemented functions from `api_coverage.py`.

### DIFF
--- a/nums/numpy/api.py
+++ b/nums/numpy/api.py
@@ -382,10 +382,22 @@ def where(condition, x=None, y=None):
     return _instance().where(condition)
 
 
-def all(a: BlockArray, axis=None, out=None, keepdims=False):
+def all(a: BlockArray, axis=None, out=None, keepdims=False, dtype=None):
     if out is not None:
         raise NotImplementedError("'out' is currently not supported.")
-    return _instance().all(a, axis=axis, keepdims=keepdims)
+    return _instance().reduce("all", a, axis=axis, keepdims=keepdims, dtype=dtype)
+
+
+def alltrue(a: BlockArray, axis=None, out=None, dtype=None, keepdims=False):
+    if out is not None:
+        raise NotImplementedError("'out' is currently not supported.")
+    return _instance().reduce("alltrue", a, axis=axis, keepdims=keepdims, dtype=dtype)
+
+
+def any(a: BlockArray, axis=None, out=None, keepdims=False, dtype=None):
+    if out is not None:
+        raise NotImplementedError("'out' is currently not supported.")
+    return _instance().reduce("any", a, axis=axis, keepdims=keepdims, dtype=dtype)
 
 ############################################
 # Utility Ops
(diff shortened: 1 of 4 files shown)


MohamedElgharbawy:
Looking into the Travis errors.

tests/numpy/test_misc.py
```diff
@@ -121,12 +121,102 @@ def test_all(nps_app_inst):
     assert nps.all(false_float).dtype is np.bool
 
 
+def test_alltrue(nps_app_inst):
```
elibol, line 124: Is `alltrue` the same as `all`? If so, I think it's okay to test `all` and `alltrue` in the same test.

tests/numpy/test_misc.py
```diff
@@ -121,12 +121,102 @@ def test_all(nps_app_inst):
...
+
+    assert nps.alltrue(true_int).dtype is np.bool
+    assert nps.alltrue(false_int).dtype is np.bool
+    assert nps.alltrue(true_bool).dtype is np.bool
+    assert nps.alltrue(false_bool).dtype is np.bool
+    assert nps.alltrue(true_float).dtype is np.bool
+    assert nps.alltrue(false_float).dtype is np.bool
+
+
+def test_any(nps_app_inst):
```
elibol, line 152: This looks very similar to `all` and `alltrue`. Can we consolidate the code for these tests?

elibol (commented):
Looks good, and thanks for cleaning up `api_coverage.py`! Please see comments.

tests/numpy/test_misc.py
MohamedElgharbawy: Yes, I think `all` replaced `alltrue` in NumPy, but both are still supported. I'll combine their tests.

tests/numpy/test_misc.py
MohamedElgharbawy: Yep!

nums/numpy/api.py
```diff
@@ -398,6 +410,11 @@ def allclose(a: BlockArray, b: BlockArray, rtol=1.e-5, atol=1.e-8, equal_nan=Fal
     return _instance().allclose(a, b, rtol, atol)
 
 
+def array_equal(a1: BlockArray, a2: BlockArray, equal_nan=False) -> BlockArray:
```
elibol, line 413: I think this implementation is okay, but please test the semantics of `array_equal` rigorously to ensure that `allclose` with 0 tolerance is an equivalent function to `array_equal`. Alternatively, we can discuss how to generalize the `allclose` function in application to support both `allclose` and `array_equal`.

MohamedElgharbawy: `allclose` uses the following equation: `absolute(a - b) <= (atol + rtol * absolute(b))`. It's typically used for float arrays to prevent floating point error, while `array_equal` is meant for integer arrays. Keeping those usage cases in mind, I believe this implementation should be okay, right? I have a test in `test_array_equal` that has a very small difference in floating points, and `array_equal` returns false as expected.

elibol: Okay. To be safe, please test edge cases such as positive/negative zero, behavior for pos/neg infinity, and comparisons between every combination of bool, int, float, and complex types (e.g. is `False` equal to `0.0`).

MohamedElgharbawy: Will do.

PR #15, opened 2021-01-25
Implemented alltrue, any

MohamedElgharbawy:
Also updated `api_coverage.py` with implemented methods.

### DIFF
--- a/nums/numpy/api.py
+++ b/nums/numpy/api.py
@@ -385,7 +385,19 @@ def where(condition, x=None, y=None):
 def all(a: BlockArray, axis=None, out=None, keepdims=False):
     if out is not None:
         raise NotImplementedError("'out' is currently not supported.")
-    return _instance().all(a, axis=axis, keepdims=keepdims)
+    return _instance().reduce("all", a, axis=axis, keepdims=keepdims)
+
+
+def alltrue(a: BlockArray, axis=None, out=None, dtype=None, keepdims=False):
+    if out is not None:
+        raise NotImplementedError("'out' is currently not supported.")
+    return _instance().reduce("alltrue", a, axis=axis, keepdims=keepdims, dtype=dtype)
+
+
+def any(a: BlockArray, axis=None, out=None, keepdims=False):
+    if out is not None:
+        raise NotImplementedError("'out' is currently not supported.")
+    return _instance().reduce("any", a, axis=axis, keepdims=keepdims)
 
 ############################################
 # Utility Ops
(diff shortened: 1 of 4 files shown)


tests/numpy/test_misc.py
```diff
@@ -113,20 +113,55 @@ def test_all(nps_app_inst):
...
+    false_float = nps.array([[0.0, 0.0, 0.0], [0.0, 0.0, 0.0]]).reshape(block_shape=(2, 2))
+
+    assert nps.any(true_int)
+    assert not nps.any(false_int)
+    assert nps.any(true_bool)
+    assert not nps.any(false_bool)
+    assert nps.any(true_float).get() == np.any(np.array([[1.0, 0.0, 0.0], [0.0, 0.0, 0.0]]))
+    assert nps.any(false_float).get() == np.any(np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 0.0]]))
+
     assert nps.all(true_int).dtype is np.bool
```
elibol, line 136: Can you put `true_int`, `false_int`, etc. in a list and run these assertions in a loop?

tests/numpy/test_misc.py
```diff
@@ -113,20 +113,55 @@ def test_all(nps_app_inst):
...
+
+    false_int = nps.array([[0, 0, 0], [0, 0, 0]]).reshape(block_shape=(2, 2))
+    false_bool = nps.array([False, False])
+    false_float = nps.array([[0.0, 0.0, 0.0], [0.0, 0.0, 0.0]]).reshape(block_shape=(2, 2))
+
+    assert nps.any(true_int)
+    assert not nps.any(false_int)
+    assert nps.any(true_bool)
+    assert not nps.any(false_bool)
+    assert nps.any(true_float).get() == np.any(np.array([[1.0, 0.0, 0.0], [0.0, 0.0, 0.0]]))
```
elibol, line 133: Wherever possible (such as here), construct the numpy array first, and call `nps.array` on it for these kinds of tests. We'll want to make sure we're testing the function's behavior on the same arrays.

elibol (commented):
Just left a few comments in the tests.

tests/numpy/test_misc.py
MohamedElgharbawy: Definitely. Will do.

tests/numpy/test_misc.py
MohamedElgharbawy: Yep, and I'll keep doing this from now on whenever possible.

nums/numpy/api.py
```diff
@@ -382,10 +382,22 @@ def where(condition, x=None, y=None):
     return _instance().where(condition)
 
 
-def all(a: BlockArray, axis=None, out=None, keepdims=False):
+def all(a: BlockArray, axis=None, out=None, keepdims=False, dtype=None):
```
elibol, line 385: Can we remove `dtype=None`? It's not a part of the NumPy API (If someone goes from using NumS to NumPy, their code ought to work).

MohamedElgharbawy: Yep, will do.

PR #16, opened 2021-01-29
Implemented array_equal, array_equiv

MohamedElgharbawy:
Created a function `array_compare` that implements `array_equal`, `array_equiv`, and `allclose` generally.

### DIFF
--- a/nums/core/array/application.py
+++ b/nums/core/array/application.py
@@ -324,9 +324,9 @@ class ArrayApplication(object):
                                                                 grid_entry,
                                                                 grid_meta,
                                                                 syskwargs={
-                                                                     "grid_entry": grid_entry,
-                                                                     "grid_shape": grid.grid_shape
-                                                                 })
+                                                                    "grid_entry": grid_entry,
+                                                                    "grid_shape": grid.grid_shape
+                                                                })
         return rarr
 
     def concatenate(self, arrays: List, axis: int, axis_block_size: int = None):
@@ -378,7 +378,7 @@ class ArrayApplication(object):
         pos = 0
         for arr in arrays:
             delta = arr.shape[axis]
-            axis_slice = slice(pos, pos+delta)
+            axis_slice = slice(pos, pos + delta)
             result_selector = tuple([slice(None, None) for _ in range(axis)] + [axis_slice, ...])
             result_ba[result_selector] = arr
             pos += delta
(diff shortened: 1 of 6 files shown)


nums/core/systems/interfaces.py
```diff
@@ -136,7 +136,13 @@ def svd(self, arr, syskwargs: Dict):
     def inv(self, arr, syskwargs: Dict):
         raise NotImplementedError()
 
-    def allclose(self, a, b, rtol, atol, syskwargs: Dict):
+    def array_equal(self, a, b, args, syskwargs: Dict):
```
elibol, line 139: While it may make sense to break these out into separate functions in the future, for now can we consolidate these calls to a single function? Similar to how `array_eq` is implemented in `ArrayApplication`.

nums/core/array/application.py
```diff
@@ -751,17 +751,18 @@ def get(self, *arrs):
                     r.append(item)
             return r
 
-    def allclose(self, a: BlockArray, b: BlockArray, rtol=1.e-5, atol=1.e-8):
+    def array_eq(self, a: BlockArray, b: BlockArray, func_name, *args):
```
elibol, line 754: Can you update the signature for this to be `self, func_name: str, a: BlockArray, b: BlockArray, *args, **kwargs`? It's a bit more consistent with existing methods on this class.

elibol (commented):
This looks very good. I left a few comments.

nums/core/array/application.py
MohamedElgharbawy: Will do

nums/core/systems/interfaces.py
MohamedElgharbawy: Sounds good.

nums/core/array/application.py
MohamedElgharbawy: Do we need `**kwargs` in the function signature? It doesn't seem like it would be used here.

elibol: Yes, if it's not going to be used leave it out.

tests/numpy/test_misc.py
```diff
@@ -129,6 +129,48 @@ def test_all_alltrue_any(nps_app_inst):
...
+        assert nps.array_equiv(nps_array_1, nps_array_2).get() == np.array_equiv(check[0], check[1])
+        assert nps.allclose(nps_array_1, nps_array_1).get() == np.allclose(check[0], check[0])
+        assert nps.allclose(nps_array_1, nps_array_2).get() == np.allclose(check[0], check[1])
+
+        assert nps.array_equal(nps_array_1, nps_array_2).dtype is np.bool
+        assert nps.array_equiv(nps_array_1, nps_array_2).dtype is np.bool
+        assert nps.allclose(nps_array_1, nps_array_2).dtype is np.bool
+
+    # False interaction test
+    false_array_1 = np.array([[False, False], [0, 0], [0.0, 0.0]])
```
elibol, line 162: When this is declared, I believe the resulting array will all be of a single type (e.g. float64). Can you separate the different types into different arrays? bool in one, int in another, etc?

PR #22, opened 2021-02-22
Implemented nanmean

MohamedElgharbawy:
Also implemented `__invert__` (`~` operator).


tests/numpy/test_nan_reduction.py
```diff
@@ -56,7 +55,8 @@ def test_nan_reductions(nps_app_inst):
 if __name__ == "__main__":
     from nums.core import application_manager
     import nums.core.settings
-    warnings.filterwarnings('ignore', r'All-NaN (slice|axis) encountered')
+    warnings.filterwarnings("ignore", "All-NaN (slice|axis) encountered")
```
nums-project, line 58: Can you move these to the top of the module, right after the imports?

MohamedElgharbawy: Yes.

PR #25, opened 2021-02-27
Implemented nanstat and atleast<n>d functions

MohamedElgharbawy:
- Implemented: `nanstd`, `nanvar`, `atleast_1d`, `atleast_2d`, `atleast_3d`

- Fixed: `reshape`, the `shape` parameter no longer has to be a tuple

- Updated now invalid uses of `reshape(shape=<shape>)`

- Replaced some uses of the deprecated `np.bool` with `bool`

### DIFF
--- a/nums/core/array/application.py
+++ b/nums/core/array/application.py
@@ -851,7 +851,7 @@ class ArrayApplication(object):
         # If blocking is "tall-skinny," then we're done.
         if R_shape != R_block_shape:
             if reshape_output:
-                R = tsR.reshape(shape=R_shape, block_shape=R_block_shape)
+                R = tsR.reshape(R_shape, block_shape=R_block_shape)
             else:
                 R = tsR
         else:
@@ -869,9 +869,9 @@ class ArrayApplication(object):
         tsR_inverse = self.inv(tsR)
         # If blocking is "tall-skinny," then we're done.
         if R_shape != R_block_shape:
-            R_inverse = tsR_inverse.reshape(shape=R_shape, block_shape=R_block_shape)
+            R_inverse = tsR_inverse.reshape(R_shape, block_shape=R_block_shape)
             if reshape_output:
-                R = tsR.reshape(shape=R_shape, block_shape=R_block_shape)
+                R = tsR.reshape(R_shape, block_shape=R_block_shape)
             else:
                 R = tsR
         else:
@@ -961,10 +961,10 @@ class ArrayApplication(object):
         if R_shape == R_block_shape or not reshape_output:
             R = tsR
         else:
-            R = tsR.reshape(shape=R_shape, block_shape=R_block_shape)
+            R = tsR.reshape(R_shape, block_shape=R_block_shape)
 
         if Q.shape != block_shape or not reshape_output:
-            Q = Q.reshape(shape=shape, block_shape=block_shape)
+            Q = Q.reshape(shape, block_shape=block_shape)
 
         return Q, R
 
@@ -1046,7 +1046,7 @@ class ArrayApplication(object):
         Q, R = self.indirect_tsqr(X, reshape_output=False)
         R_inv = self.inv(R)
         if R_shape != R_block_shape:
-            R_inv = R_inv.reshape(shape=R_shape, block_shape=R_block_shape)
+            R_inv = R_inv.reshape(R_shape, block_shape=R_block_shape)
         theta = R_inv @ (Q.T @ y)
         return theta
 
@@ -1061,7 +1061,7 @@ class ArrayApplication(object):
         # Invert R.
         R_inv = self.inv(R)
         if R_shape != R_block_shape:
-            R_inv = R_inv.reshape(shape=R_shape, block_shape=R_block_shape)
+            R_inv = R_inv.reshape(R_shape, block_shape=R_block_shape)
         theta = R_inv @ (Q.T @ y)
         return theta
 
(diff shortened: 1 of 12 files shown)


nums/core/array/application.py
```diff
@@ -1107,3 +1107,79 @@ def nanmean(self, a: BlockArray, axis=None, keepdims=False, dtype=None):
...
+        res = []
+        for ary in arys:
+            if not isinstance(ary, BlockArray):
+                ary = np.array(ary)
+                block_shape = self.compute_block_shape(ary.shape, ary.dtype)
+                ary = self.array(ary, block_shape)
+            if ary.ndim == 0:
+                result = ary.reshape(1, 1)
+
+            # TODO (MWE): Implement this using newaxis when supported
```
elibol, line 1153: Until we have clarity on how we want to support the semantics for this case, can we simplify these implementations (from 1d to 3d) so that they all call a single, more general function? If so, can you also describe as a comment why NumPy uses newaxis?

nums/core/array/application.py
```diff
@@ -1107,3 +1107,79 @@ def nanmean(self, a: BlockArray, axis=None, keepdims=False, dtype=None):
...
+    def nanstd(self, a: BlockArray, axis=None, ddof=0, keepdims=False, dtype=None):
+        res = self.sqrt(self.nanvar(a, axis, ddof, keepdims))
+        if dtype is not None:
+            res = res.astype(dtype)
+        return res
+
+    def atleast_1d(self, *arys):
+        res = []
+        for ary in arys:
+            if not isinstance(ary, BlockArray):
```
elibol, line 1129: Can you look into using or creating a utility function out of `BlockArray.check_or_convert_other(self, other)` for this case? We can discuss details of this, but this behavior is creating a precedent that all operations auto-convert their inputs to BlockArray. We'll need to provide similar semantics for all operations.

tests/numpy/test_stack.py
```diff
@@ -0,0 +1,84 @@
...
+
+def test_atleast_1d(nps_app_inst):
+    import nums.numpy as nps
+    assert nps_app_inst is not None
+
+    x = 1.0
+    assert np.allclose(nps.atleast_1d(x).get(), np.atleast_1d(x))
+
+    x = np.arange(9).reshape(3, 3)
+    assert np.allclose(nps.atleast_1d(x).get(), np.atleast_1d(x))
```
elibol, line 31: While passing in a numpy array as-is is fine with the existing code, as I mention elsewhere, this creates a precedent that requires us to support such semantics for all operations (which is currently not true). It may make sense to provide support for auto conversion as a separate PR, with its own explicit set of tests, and explicitly convert the numpy arrays in this test to BlockArrays.

elibol (commented):
This looks really good, and I think we may need to split the auto conversion functionality you've added for the `atleast` functions into a separate PR. Supporting auto conversion for these functions means we ought to support them in all operations, for consistency. We can discuss the best course of action offline.

The stats imps look good to me :)

nums/core/array/application.py
MohamedElgharbawy: Sure.

tests/numpy/test_stack.py
MohamedElgharbawy: Sounds good.

nums/core/array/application.py
MohamedElgharbawy: On second thought, generalizing it will just have a lot of messy checks and complicated statements, I think keeping them separate is the best move.

nums/core/array/blockarray.py
```diff
@@ -125,12 +125,14 @@ def touch(self):
...
 
-    def reshape(self, shape=None, **kwargs):
+    def reshape(self, *shape, **kwargs):
         block_shape = kwargs.get("block_shape", None)
         if array_utils.is_int(shape):
             shape = (shape,)
-        elif shape is None:
+        elif len(shape) == 0:
             shape = self.shape
+        elif len(shape) == 1 and isinstance(shape[0], (tuple, list)):
```
elibol, line 134: Can you update this to be:

```
if isinstance(shape[0], (tuple, list)):
    assert len(shape) == 1
    ...
```

I'm assuming you cannot pass a list of lists or list of tuples, but please confirm by checking NumPy behavior. Also, can you add a final `else` if none of the above conditions are satisfied, checking that the `shape` provided is a list of type int?

tests/numpy/test_stack.py
```diff
@@ -0,0 +1,84 @@
...
+    assert nps_app_inst is not None
+
+    x = 1.0
+    assert np.allclose(nps.atleast_1d(x).get(), np.atleast_1d(x))
+
+    x_np = np.arange(9).reshape(3, 3)
+    x_nps = nps.arange(9).reshape(3, 3)
+    assert np.allclose(nps.atleast_1d(x_np).get(), np.atleast_1d(x_np))
+    assert np.allclose(nps.atleast_1d(x_nps).get(), np.atleast_1d(x_np))
+    assert nps.atleast_1d(x_nps) is x_nps
```
elibol, line 34: Can you test that this `is` check is also true for numpy?

tests/numpy/test_stack.py
```diff
@@ -0,0 +1,84 @@
...
+
+    x = np.arange(3.0)
+    assert np.allclose(nps.atleast_2d(x).get(), np.atleast_2d(x))
+
+    assert nps.atleast_2d(1, [1, 2], [[1, 2]])[0] == np.atleast_2d(1, [1, 2], [[1, 2]])[0]
+    assert nps.atleast_2d(1, [1, 2], [[1, 2]])[1] == np.atleast_2d(1, [1, 2], [[1, 2]])[1]
+    assert nps.atleast_2d(1, [1, 2], [[1, 2]])[2] == np.atleast_2d(1, [1, 2], [[1, 2]])[2]
+
+
+def test_atleast_3d(nps_app_inst):
```
elibol, line 55: Looks like you test a scalar, a 1d, 2d, and 3d array passed to the function for scalars and lists (Python objects). From our discussion offline, there are 3 types we want to test: Python objects, numpy objects and BlockArray objects. In addition, it seems to properly test this functions behavior, we want to test a scalar, and data structures with 1, 2, 3, and 4 axes for each of these different types, and a few mixtures of those types as well. I think you can test these cases more easily by creating a function that generates the test inputs up to a certain dimension. Could you look into doing this for this test, as well as the other `atleast_nd` tests?

nums/core/array/blockarray.py
MohamedElgharbawy: Will do. I can confirm you can't pass in a list of lists/tuples.

tests/numpy/test_stack.py
MohamedElgharbawy: Yep.

tests/numpy/test_stack.py
MohamedElgharbawy: Will do.

tests/numpy/test_stack.py
```diff
@@ -0,0 +1,98 @@
...
+
+    scalar = 1
+    tuple_1d = (1, 2)
+    list_1d = [1, 2]
+    np_2d = np.array([[1, 2]])
+    np_3d = np.array([[[1, 2]]])
+    nps_3d = nps.array([[[1, 2]]])
+    list_4d = [[[[1, 2]]]]
+
+    if n == 1:
```
elibol, line 33: Can we simplify this code by passing the np and nps functions to compare?

tests/numpy/test_stack.py
```diff
@@ -0,0 +1,98 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import numpy as np
+
+
+# pylint: disable=import-outside-toplevel
+
+def try_multiple_nd(n):
```
elibol, line 22: This looks good, and we can keep it as-is, but would be great to improve coverage by doing something like this:

```python
import itertools

python_types = [
1, 
[1, 2], 
[[1,2]],
[[[1,2]]]
]
all_types = python_types + list(map(np.array, all_types))
test_cases = list(itertools.product(all_types, repeat=3))

for arguments in test_cases:
   if n == 1:
      np_res = np.atleast_1d(*arguments)
      nps_res = nps.atleast_1d(*arguments)
      ...
```

This should cover all the numpy cases, and I think it's okay given the small number of cases it ends up generating. It would also be good to mix in a few random cases with BlockArray instances for the `nps` implementations, e.g. by drawing from a set like this:

```python
ba_cases = list(map(nps.array, all_types))
```

nums/core/array/application.py
```diff
@@ -1116,3 +1116,86 @@ def nanmean(self, a: BlockArray, axis=None, keepdims=False, dtype=None):
         if dtype is not None:
             res = res.astype(dtype)
         return res
+
+    def nanvar(self, a: BlockArray, axis=None, ddof=0, keepdims=False, dtype=None):
```
elibol, line 1120: Do we have test coverage for nanvar and nanstd?

elibol (commented):
Looking good! Left some comments.

tests/numpy/test_stack.py
MohamedElgharbawy: Yes.

MohamedElgharbawy: Yes.

tests/numpy/test_stack.py
MohamedElgharbawy: ```all_types = python_types + list(map(np.array, all_types))```

Is the second parameter of `map` correct? How can it use `all_types` before it's created?

nums/core/array/application.py
MohamedElgharbawy: Yes, in `test_nan_reductions`.

tests/numpy/test_stack.py
elibol: Yep you're right, the input to map should be `python_types`.

PR #27, opened 2021-03-11
Add NotImplementedError for matmul with tensors

briancpark:
Currently, there is a correctness error with matmul and tensors of rank > 2. Disabled the current implementation, and added a NotImplementedError for now.

As stated in the [NumPy docs](https://numpy.org/doc/stable/reference/generated/numpy.matmul.html), the correct behavior should be:

> If either argument is N-D, N > 2, it is treated as a stack of matrices residing in the last two indexes and broadcast accordingly.

NumS is calling tensordot instead, and a separate subroutine for matmul with tensors needs to be implemented.

### DIFF
--- a/nums/core/array/blockarray.py
+++ b/nums/core/array/blockarray.py
@@ -410,7 +410,8 @@ class BlockArray(BlockArrayBase):
 
     def __matmul__(self, other):
         if len(self.shape) > 2:
-            return self.tensordot(other, 2)
+            # TODO: (bcp) NumPy's implementation does a stacked matmul, which is not supported yet.
+            raise NotImplementedError("Matrix multiply for tensors of rank > 2 not supported yet")
         else:
             return self.tensordot(other, 1)
 
--- a/nums/core/array/blockarray.py
+++ b/nums/core/array/blockarray.py
@@ -411,7 +411,7 @@ class BlockArray(BlockArrayBase):
     def __matmul__(self, other):
         if len(self.shape) > 2:
             # TODO: (bcp) NumPy's implementation does a stacked matmul, which is not supported yet.
-            raise NotImplementedError("Matrix multiply for tensors of rank > 2 not supported yet")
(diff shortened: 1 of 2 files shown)


tests/numpy/test_arithmetic.py
```diff
@@ -114,6 +114,7 @@ def check_tensordot_op(_np_a, _np_b, axes):
     np_B = np.transpose(np.stack([np_w + i for i in range(20)]))
     check_matmul_op(np_A, np_B)
 
+# TODO(bcp): Add matmul tests for rank > 2
```
elibol, line 117: (nit) Missing punctuation in this TODO.

nums/core/array/blockarray.py
```diff
@@ -410,7 +410,8 @@ def reduce_axis(self, op_name, axis, keepdims=False):
 
     def __matmul__(self, other):
         if len(self.shape) > 2:
-            return self.tensordot(other, 2)
+            # TODO: (bcp) NumPy's implementation does a stacked matmul, which is not supported yet.
+            raise NotImplementedError("Matrix multiply for tensors of rank > 2 not supported yet")
```
elibol, line 414: (nit) missing punctuation in error message.

elibol (commented):
Can you add a test in `tests/numpy/array/test_arithmetic.py` that tests this error? You can do this with pytest using

```
with pytest.raises(NotImplementedError):
    C: BlockArray = A @ B
```

Where `A` and `B` are matrices that will trigger the error.

briancpark:
Done! For some reason when calling:
```python
with pytest.raises(NotImplementedError):
     C: BlockArray = A @ B
```
It fails the pylint test with `Statement seems to have no effect (pointless-statement)`.

I looked more deeply into it, and it requires a `touch()` call, and I'm not sure why the `@` operator acts like an asynchronous call while `matmul()` works fine?

elibol:
I agree that's unusual error. You could alternatively tie it all together with an assertion, but this looks good as-is, so I'll go ahead and merge :)

PR #28, opened 2021-03-21
Swapaxes API

kssteven418:
`swapaxes` API added, and test code added

### DIFF
--- a/nums/numpy/api.py
+++ b/nums/numpy/api.py
@@ -330,6 +330,9 @@ def squeeze(a: BlockArray, axis=None):
     assert axis is None, "axis not supported."
     return a.squeeze()
 
+def swapaxes(x: BlockArray, axis1: int, axis2: int):
+    return x.swapaxes(axis1, axis2)
+
 
 def transpose(a: BlockArray, axes=None):
     assert axes is None, "axes not supported."
--- a/nums/numpy/api.py
+++ b/nums/numpy/api.py
@@ -330,7 +330,7 @@ def squeeze(a: BlockArray, axis=None):
     assert axis is None, "axis not supported."
     return a.squeeze()
 
-def swapaxes(x: BlockArray, axis1: int, axis2: int):
+def swapaxes(a: BlockArray, axis1: int, axis2: int):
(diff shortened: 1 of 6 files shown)


nums/numpy/api.py
```diff
@@ -330,6 +330,9 @@ def squeeze(a: BlockArray, axis=None):
     assert axis is None, "axis not supported."
     return a.squeeze()
 
+def swapaxes(x: BlockArray, axis1: int, axis2: int):
```
elibol, line 333: Can we change the name of the input array to be `a` instead of `x`? This is the variable name in numpy, in case keyword arguments are used.

tests/numpy/test_creation_manipulation.py
```diff
@@ -162,6 +188,14 @@ def check_dim(_np_a, _ns_a):
     check_expand_and_squeeze(np_A, axes=(0, 5, 6))
     check_expand_and_squeeze(np_A, axes=(2, 3, 5, 6, 7))
 
+    check_swapaxes(np_A, axis1=0, axis2=1)
+    check_swapaxes(np_A, axis1=0, axis2=2)
+    check_swapaxes(np_A, axis1=0, axis2=3)
+    check_swapaxes(np_A, axis1=1, axis2=2)
+    check_swapaxes(np_A, axis1=1, axis2=3)
+    check_swapaxes(np_A, axis1=2, axis2=3)
+    check_swapaxes(np_A, axis1=3, axis2=3)
```
elibol, line 197: Wondering, why test `axis1 == axis2 == 3` only and not 0, 1, and 2?

elibol (commented):
Left very minor comments. Great job with this. It is clean and should perform well.

nums/numpy/api.py
kssteven418: I've changed the name

tests/numpy/test_creation_manipulation.py
kssteven418: There's no specific reason,  just wanted to make sure that the case in which `axis1` and `axis2` are the same works well. I just changed the code to test all the possible combinations.

PR #29, opened 2021-03-22
Trace implemented

Priyansdesai:
Added:
1. Trace implementation
2. Test cases for trace
3. Added some changes in Readme for Windows support


MohamedElgharbawy:
Can you squash this into one commit?

README.md
```diff
@@ -208,6 +208,18 @@ source activate nums
 pip install -e ".[testing]"
 ```
 
+For Windows, follow the steps below if you do not have Anaconda already downloaded. 
+1. Follow the steps to download the Anaconda Installer. https://docs.anaconda.com/anaconda/install/windows/
+2. During the setup, select that you want Anaconda to be added to the PATH variable. 
+3. Fork/Clone the nums-project repository
+4. Open Command Prompt and follow the instructions below:
```
elibol, line 215: Can you update the README so that the #Installation section has two subsections: One for Linux/Mac and one for Windows, and copy and adapt these instructions for that section? It's okay if there are redundant instructions between the two subsections.

Can you update these instructions to read as follows:
**For Windows, we recommend Anaconda. Download and install Anaconda for Windows [here](https://docs.anaconda.com/anaconda/install/windows/). Make sure to add Anaconda to the PATH variable during installation.**

README.md
```diff
@@ -208,6 +208,18 @@ source activate nums
...
+    ```sh
```
elibol, line 216: Can you remove this sh block and just update the one above from `source activate nums` to `conda activate nums`?

tests/numpy/test_creation_manipulation.py
```diff
@@ -78,6 +79,37 @@ def test_diag(nps_app_inst):
...
 
+def test_trace(nps_app_inst):
+    import nums.numpy as nps
+
+    assert nps_app_inst is not None
+
+    a: BlockArray = nps.array([1.0, 2.0, 3.0, 4.0])
+
+    # Test that trace doesn't work with 1-dimensional vectors
+    with pytest.raises(ValueError):
```
elibol, line 90: This is good. Can you add another test here for dtype? Not as important (it is easy to see that it's correct), but might be worth also testing errors for arrays with 3+ axes, the offset, axis1, and axis2 params.

tests/numpy/test_creation_manipulation.py
```diff
@@ -78,6 +79,37 @@ def test_diag(nps_app_inst):
     np_arr = np.diag(np_arr)
     assert np.allclose(ba.get(), np_arr)
 
+def test_trace(nps_app_inst):
```
elibol, line 82: Can you add a test for dtype?

tests/numpy/test_creation_manipulation.py
```diff
@@ -78,6 +79,37 @@ def test_diag(nps_app_inst):
...
+    a_diag_np = np.diag(a.get()) # Numpy version
+
+    # Apply on Trace on each of the Diagonal matrices
+    a_diag_trace = nps.trace(a_diag).get()
+    a_diag_np_trace = np.trace(a_diag_np)
+
+    assert np.allclose(a_diag_trace, a_diag_np_trace)
+
+    # Use pre-created diagonal matrices
+    b: BlockArray = nps.array([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0], [7.0, 8.0, 9.0]])
```
elibol, line 104: Can you test non-square matrices (short-wide and tall-skinny)?

elibol (commented):
This is looking good :) Left a few comments.

You can find the linting issues here: https://travis-ci.com/github/nums-project/nums/builds/220765545#L574-L581

To fix them locally, run `pytest --pylint`.

tests/numpy/test_creation_manipulation.py
Priyansdesai: The behavior of NumPy and NumS is different for diagonal matrices. Doing the following 
`>>> a = np.array([[1, 2], [3,4], [5,6]])
>>> np.diag(a)
array([1, 4])
`
But doing the equivalent in NumS, produces an assertion error. 
`>>> a: BlockArray = nps.array([[1, 2], [3, 4], [5, 6]])
>>> nps.diag(a)
AssertionError`

This is triggered by the line `assert X.shape[0] == X.shape[1]` from Line 423 in core/array/application.py, enforcing that X needs to be square.

tests/numpy/test_creation_manipulation.py
Priyansdesai: I have added a test for this, wherein the elements are like 1.01, 2.02, 3.03 and 4.04 and I need an int sum.

tests/numpy/test_creation_manipulation.py
Priyansdesai: Done

README.md
Priyansdesai: So, do you still want two different blocks of instructions - one for Linux/Mac and one for Windows, as per comment above.

elibol: Yes, create a separate section in "Installation" for Windows, and let's see what it looks like. If it's just a couple of sentences, then we can simplify.

In this section ("Contributing") it doesn't look like we need a separate Windows-based set of instructions.

Alternatively, I can help with this part of the PR based on what you've written thus far if you give me push permissions to your repo.

tests/numpy/test_creation_manipulation.py
elibol: Indeed, let's discuss offline further to decide whether we want to work on diag to support these cases. I can help with the diag imp if needed.

README.md
Priyansdesai: Okay sure. I have given you access, I think you have a better idea of how the final set of instructions should work.

nums/core/array/application.py
```diff
@@ -425,7 +425,7 @@ def diag(self, X: BlockArray) -> BlockArray:
                                                                         grid_meta,
                                                                         syskwargs=syskwargs)
         elif len(X.shape) == 2:
-            assert X.shape[0] == X.shape[1]
+            assert X.shape[0] == X.shape[1], "X must be a square matrix"
```
elibol, line 428: Can you include punctuation? Can you also add a similar error message to the `block_shape` assertion below?

tests/test_text.out
```diff
@@ -0,0 +1,34 @@
+# field1,field2,field3
```
elibol, line 1: Was this mistakenly committed?

tests/numpy/test_creation_manipulation.py
```diff
@@ -78,6 +79,55 @@ def test_diag(nps_app_inst):
...
+    with pytest.raises(NotImplementedError):
+        nps.trace(mat_diag, offset=2)
+
+    # Test data type of the return type
+    mat_diag = nps.diag(nps.array([1.01, 2.02, 3.03, 4.04]))
+    mat_diag_np = np.diag(np.array([1.01, 2.02, 3.03, 4.04]))
+    mat_diag_trace = nps.trace(mat_diag, dtype=int).get()
+    mat_diag_np_trace = np.trace(mat_diag_np, dtype=int)
+
+    assert np.allclose(mat_diag_trace, mat_diag_np_trace)
```
elibol, line 129: This is good. Can we include a test that ensures that, if the input dtype is integer, the output dtype is also integer?

elibol (commented):
This is looking great. I left a few comments.

nums/core/array/application.py
Priyansdesai: Sure. You mean punctuation in the error message right?

tests/test_text.out
Priyansdesai: Yes. My bad. Will delete it. Sorry about that!

tests/numpy/test_creation_manipulation.py
Priyansdesai: Sure. Let me try that. Probably check the datatype of the first element or just find a way to get the type of the whole array.

PR #30, opened 2021-03-23
Implemented average, formatting in README

dlzou:
`average` currently does not support a tuple of axes due to `reduce_axis`. It also currently only supports `a` and `weights` having the same shape. I have tested locally for correctness in serial mode.
The change in README is to fix a small rendering issue in the "Contributing" section.


nums/numpy/api.py
```diff
@@ -444,6 +446,29 @@ def any(a: BlockArray, axis=None, out=None, keepdims=False):
...
+            returned: bool = False):
+    if weights is None:
+        avg = mean(a, axis)
+        weights_sum = BlockArray.from_scalar(a.size / avg.size, a.system)
+    else:
+        if a.shape != weights.shape:
+            raise NotImplementedError("'average' currently does not support broadcasting;"
+                                      "dimensions of 'a' and 'weights' must match.")
+        weights_sum = sum(weights, axis=axis)
+        if not all(weights_sum):
```
elibol, line 461: Just want to check: Is this what NumPy raises if there are zeros in some entries of `weights_sum`? Can you test that this error is raised when there are zeros?

nums/numpy/api.py
```diff
@@ -444,6 +446,29 @@ def any(a: BlockArray, axis=None, out=None, keepdims=False):
     return _instance().reduce("any", a, axis=axis, keepdims=keepdims)
 
 
+def average(a: BlockArray,
+            axis: Union[None, int] = None,
+            weights: Union[None, BlockArray] = None,
+            returned: bool = False):
+    if weights is None:
+        avg = mean(a, axis)
+        weights_sum = BlockArray.from_scalar(a.size / avg.size, a.system)
```
elibol, line 455: Any clean way to compute `weights_sum` here only if it's returned? It's very low overhead. It's okay to keep as-is to maintain readability.

tests/numpy/test_np_reduction.py
```diff
@@ -85,10 +85,77 @@ def test_argops(nps_app_inst):
...
+                    ba_avg, ba_ws = ba_result[0], ba_result[1]
+                    assert np.allclose(ba_ws.get(), np_ws)
+                    assert ba_avg.grid.grid_shape == ba_avg.blocks.shape
+                    assert ba_avg.shape == np_avg.shape
+                    assert np.allclose(ba_avg.get(), np_avg)
+                else:
+                    assert ba_result.grid.grid_shape == ba_result.blocks.shape
+                    assert ba_result.shape == np_result.shape
+                    assert np.allclose(ba_result.get(), np_result)
+
```
elibol, line 152: Just repeating here: Would be nice to add some tests that would trigger zeros in the weight sum. Doesn't need to be a part of the above loop.

elibol (commented):
Very nice implementation. I left a few minor comments.

dlzou:
Once `swapaxes` is merged I can add the auto-broadcasting of `weights` when it is 1D.

nums/numpy/api.py
dlzou: Yes, NumPy source code checks if `any` element is equal to zero. Will add tests to show this.

elibol:
Looks good to me.

MohamedElgharbawy:
Can you squash this into one commit?

PR #32, opened 2021-03-27
Shape mismatch error/assertion checks for tensordot

briancpark:
Added synchronous error checking for tensordot to make it less confusing to the programmer where the error happens. Before, calling tensordot with mismatched dimensions would only be caught after operations were passed into Ray. 

The code for error checking was modified from the [NumPy source code](https://github.com/numpy/numpy/blob/v1.20.0/numpy/core/numeric.py#L949-L1139). This should make it easier for the programmer to catch errors, as well as catching errors in our own unit tests without the need of calling `.touch()`.


briancpark:
Hmm. I'm actually confused what the bug is and why the Travis check failed? On my end, this is what I'm getting locally (posting this because I don't think Travis is showing which tests failed):
```
======================================================================================================================== short test summary info ========================================================================================================================
FAILED nums/core/array/blockarray.py::/Users/brianpark/Desktop/NumS/fork/nums/core/array/blockarray.py::PYLINT
FAILED tests/numpy/test_arithmetic.py::/Users/brianpark/Desktop/NumS/fork/tests/numpy/test_arithmetic.py::PYLINT
FAILED tests/numpy/test_arithmetic.py::test_inner_outer[serial] - NotImplementedError: NumPy function <function inner at 0x7fd5d001cc20> not yet implemented.
FAILED tests/numpy/test_arithmetic.py::test_inner_outer[ray-cyclic] - NotImplementedError: NumPy function <function inner at 0x7fd5d001cc20> not yet implemented.
FAILED tests/numpy/test_misc.py::test_properties[serial] - NotImplementedError: NumPy function <function shape at 0x7fd61011fdd0> not yet implemented.
FAILED tests/numpy/test_misc.py::test_properties[ray-cyclic] - NotImplementedError: NumPy function <function shape at 0x7fd61011fdd0> not yet implemented.
```

briancpark:
Disregard the previous comment. I think my local environment was set up wrong when testing/

nums/core/array/blockarray.py
```diff
@@ -410,14 +410,59 @@ def reduce_axis(self, op_name, axis, keepdims=False):
 
     def __matmul__(self, other):
         if len(self.shape) > 2:
-            return self.tensordot(other, 2)
+            # TODO: (bcp) NumPy's implementation does a stacked matmul, which is not supported yet.
```
elibol, line 413: Can you rewrite this as `# TODO (bcp): ...`

nums/core/array/blockarray.py
```diff
@@ -410,14 +410,59 @@ def reduce_axis(self, op_name, axis, keepdims=False):
...
 
     def tensordot(self, other, axes=2):
         if not isinstance(other, BlockArray):
             raise ValueError("Cannot automatically construct BlockArray for tensor operations.")
 
+        # Error checking before everything gets passed into BlockArray operations. Modified from the
+        # original NumPy tensordot method for error checking:
+        # https://github.com/numpy/numpy/blob/v1.20.0/numpy/core/numeric.py#L949-L1139
+        try:
+            iter(axes)
```
elibol, line 426: Can we move this entire try/except sequence of tests into a function in `nums/core/array/utils.py`? Pass in the parameters needed to run the test, and name it `np_tensordot_param_test`.

Call `np_tensordot_param_test` from here, and before calling `np_tensordot_param_test`, test to ensure `axes` is integer, because currently that's all we support.

tests/numpy/test_arithmetic.py
```diff
@@ -126,9 +144,48 @@ def test_inner_outer(nps_app_inst):
...
+        for _op in _ops:
+            ns_op = nps.__getattribute__(_op)
+            np_op = np.__getattribute__(_op)
+            _ns_a = nps.array(_np_a)
+            _ns_b = nps.array(_np_b)
+
+            with pytest.raises(ValueError):
+                np_op(_np_a, _np_b)
+            with pytest.raises(ValueError):
+                # TODO(bcp): Add synchronous broadcast checks for simple operands.
```
elibol, line 165: Add a space between TODO and your initials, so the TODO reads `# TODO (bcp): ...`

tests/numpy/test_arithmetic.py
```diff
@@ -126,9 +144,48 @@ def test_inner_outer(nps_app_inst):
     assert np.allclose(np.inner(A, B), nps.inner(nps_A, nps_B).get())
     assert np.allclose(np.outer(A, B), nps.outer(nps_A, nps_B).get())
 
+def test_broadcast(nps_app_inst):
```
elibol, line 147: This is good, but I think this is missing coverage for all variations of the parameter `axes`. Can you add tests that cover the different types of `axes` we don't support?

Since this is an error test, can you rename the test to `test_broadcast_error`?

tests/numpy/test_arithmetic.py
```diff
@@ -126,9 +144,48 @@ def test_inner_outer(nps_app_inst):
...
+    import numpy as np
+    from nums import numpy as nps
+    import pytest
+    assert nps_app_inst is not None
+
+    def check_matrix_broadcast_mismatch_simple(_np_a, _np_b):
```
elibol, line 153: Can we rename this to `check_broadcast_mismatch_simple`? The `matrix` in the function name suggests there's a matrix multiply being tested, but it's element-wise ops on 2d arrays.

tests/numpy/test_arithmetic.py
```diff
@@ -126,9 +144,48 @@ def test_inner_outer(nps_app_inst):
...
+            _ns_b = nps.array(_np_b)
+
+            with pytest.raises(ValueError):
+                np_op(_np_a, _np_b)
+            with pytest.raises(ValueError):
+                # TODO(bcp): Add synchronous broadcast checks for simple operands.
+                res = ns_op(_ns_a, _ns_b)
+                res.touch()
+
+    def check_matrix_broadcast_mismatch_tensor(_np_a, _np_b, axes):
```
elibol, line 169: Same as above, can we remove the `matrix` in func name?

tests/numpy/test_arithmetic.py
```diff
@@ -115,6 +115,24 @@ def check_tensordot_op(_np_a, _np_b, axes):
     check_matmul_op(np_A, np_B)
 
 
+def test_matmul_tensor(nps_app_inst):
```
elibol, line 118: Can you rename this to `test_matmul_tensor_error`?

tests/numpy/test_arithmetic.py
```diff
@@ -115,6 +115,24 @@ def check_tensordot_op(_np_a, _np_b, axes):
...
+
+    from nums import numpy as nps
+
+    import pytest
+
+    assert nps_app_inst is not None
+
+    # TODO(bcp): Replace with matmul tests for rank > 2 once implemented.
```
elibol, line 126: Update TODO comment to look like others.

elibol (commented):
This is looking good. Left a few comments.

briancpark:
Please let me know if the tensordot needs more coverage than what I pushed. Also, looking more into it, I think we can add a quick shape/broadcast error check for `map_bop()` for all binary operations under `applications.py`. I can also add on to this PR if you want. Just let me know if you would prefer that error check under `utils.py`, because it should only take two lines as such:
```python
if arr_1.shape != arr_2.shape:
    raise ValueError
```

elibol:
> Please let me know if the tensordot needs more coverage than what I pushed. Also, looking more into it, I think we can add a quick shape/broadcast error check for `map_bop()` for all binary operations under `applications.py`. I can also add on to this PR if you want. Just let me know if you would prefer that error check under `utils.py`, because it should only take two lines as such:
> 
> ```python
> if arr_1.shape != arr_2.shape:
>     raise ValueError
> ```

Will go over this again a little later and merge if all looks good.

For binary ops, it may be a bit more complicated. See  #33.

nums/core/array/blockarray.py
```diff
@@ -445,6 +445,9 @@ def tensordot(self, other, axes=2):
         if not isinstance(other, BlockArray):
             raise ValueError("Cannot automatically construct BlockArray for tensor operations.")
 
+        assert isinstance(axes, int)
```
elibol, line 448: Can you test this using an `if` statement, and raise an appropriate error if the test fails?

tests/numpy/test_arithmetic.py
```diff
@@ -123,14 +123,14 @@ def test_matmul_tensor(nps_app_inst):
 
     assert nps_app_inst is not None
 
-    # TODO(bcp): Replace with matmul tests for rank > 2 once implemented.
-    def check_matmul_tensor(_ns_a, _ns_b):
+    # TODO (bcp): Replace with matmul tests for rank > 2 once implemented.
+    def test_matmul_tensor_error(_ns_a, _ns_b):
```
elibol, line 127: Can we revert the prefix of this function back to `check`? The `test` prefix is used to designate test functions to pytest. While this is a nested function, it will improve readability to have the prefix be something other than `test`.

elibol (commented):
Looks good so far. Left minor comments. There's still one comment from my previous review that needs to be addressed. I think it's minor.

nums/core/array/blockarray.py
```diff
@@ -445,6 +445,11 @@ def tensordot(self, other, axes=2):
         if not isinstance(other, BlockArray):
             raise ValueError("Cannot automatically construct BlockArray for tensor operations.")
 
+        if not isinstance(axes, int):
+            raise TypeError(f"'{type(axes).__name__}' object cannot be interpreted as an integer")
```
elibol, line 449: The `tensordot` supports non-integer axes; could be a tuple or array like structure of integers. Can we update this check as follows:

1. If it's an integer, ok.
2. If it's "array like", raise `NotImplementedError`. You can use `nums.core.array.utils.is_array_like`.
3. If it's neither an int or array like, then raise a type error, indicating that an integer or array like was expected, but got type(axes) instead.

nums/core/array/blockarray.py
```diff
@@ -445,6 +445,11 @@ def tensordot(self, other, axes=2):
...
+
+        array_utils.np_tensordot_param_test(self.shape, self.ndim, other.shape, other.ndim, axes)
```
elibol, line 451: Can you rewrite this so that the `array_utils.np_tensordot_param_test` returns a bool, and we raise the error in this function?

tests/numpy/test_arithmetic.py
```diff
@@ -145,7 +145,69 @@ def test_inner_outer(nps_app_inst):
     assert np.allclose(np.outer(A, B), nps.outer(nps_A, nps_B).get())
 
 
-# TODO(hme): Add broadcast tests.
+def test_broadcast_error(nps_app_inst):
```
elibol, line 148: We're testing for valid tensordot shapes, which is different than broadcasting. The broadcasting tests ought to be for element-wise operations (and matmul for stacked matrices I believe). Can we revive the broadcast tests TODO, and rename this test to something like `test_tensordot_shape_errors` (and something consistent with the `check_` functions below)?

If you want to add broadcast tests, would be best to add those as a separate PR.

elibol (commented):
Looks good. Very close on this one. Left some comments.

nums/core/array/blockarray.py
```diff
@@ -484,6 +484,20 @@ def _compute_tensordot_syskwargs(self, self_block: Block, other_block: Block):
                 return other_block.grid_entry, other_block.grid_shape
 
     def tensordot(self, other, axes=2):
+        if not isinstance(other, BlockArray):
+            raise ValueError("Cannot automatically construct BlockArray for tensor operations.")
+
+        if isinstance(axes, int):
+            pass
+        elif array_utils.is_array_like(axes):
+            raise NotImplementedError("Tensordot across multiple axes not supported yet.")
```
elibol, line 493: Let's update this to say "non-integer axes is currently not supported."

nums/core/array/blockarray.py
elibol: For the type error, can we update this to say f"Unexpected axes type '{type(axes).__name__}'"

elibol (commented):
Left a couple of very minor comments.

tests/numpy/test_arithmetic.py
```diff
@@ -144,9 +144,78 @@ def test_inner_outer(nps_app_inst):
...
 
-# TODO(hme): Add broadcast tests.
+def test_tensordot_shape_error(nps_app_inst):
+    import numpy as np
+    from nums.core.array.utils import is_array_like
+    from nums import numpy as nps
+    import pytest
+    assert nps_app_inst is not None
 
+    def check_simple_shape_mismatch_error(_np_a, _np_b):
```
elibol, line 156: Do you remember why we added this? Since this is not testing tensordot operations, I think we can remove this.

tests/numpy/test_arithmetic.py
```diff
@@ -144,9 +144,78 @@ def test_inner_outer(nps_app_inst):
...
     assert np.allclose(np.outer(A, B), nps.outer(nps_A, nps_B).get())
 
+# TODO (hme): Add broadcast tests.
 
-# TODO(hme): Add broadcast tests.
+def test_tensordot_shape_error(nps_app_inst):
+    import numpy as np
+    from nums.core.array.utils import is_array_like
+    from nums import numpy as nps
+    import pytest
```
elibol, line 153: Can you move numpy and pytest imports to the top of the file?

elibol (commented):
Just left a couple of minor comments.

tests/numpy/test_arithmetic.py
briancpark: I've also went ahead and moved all the nums imports to the top as well. It should look much neater and consistent now.

PR #34, opened 2021-04-03
Added fallback functionality

MohamedElgharbawy:
- Moved `fallback` from `api_coverage.py` to `settings.py`
- Fixed small bugs in various methods that were found in failed doctests
- Updated deprecated uses of `np.bool`, `np.float`, and `np.int`
- Fixed `arange` not being able to handle float `start` and `stop` indices
- Fixed `linspace` normalizing incorrectly following the `arange` fix
- Removed the [NumPy Financial functions](https://www.w3resource.com/numpy/financial-functions/index.php) from the list of fallback methods
- Added a warning to fallback functions.

### DIFF
--- a/nums/api.py
+++ b/nums/api.py
@@ -14,8 +14,6 @@
 # limitations under the License.
 
 
-import numpy as np
-
 from nums.core.application_manager import instance as _instance
 from nums.core.array.blockarray import BlockArray
 
@@ -68,7 +66,7 @@ def delete(filename: str) -> BlockArray:
         return _instance().delete_fs(filename)
 
 
-def read_csv(filename, dtype=np.float, delimiter=',', has_header=False) -> BlockArray:
+def read_csv(filename, dtype=float, delimiter=',', has_header=False) -> BlockArray:
     """Read a csv text file.
 
     Args:
--- a/nums/core/array/application.py
+++ b/nums/core/array/application.py
@@ -106,7 +106,7 @@ class ArrayApplication(object):
                     cluster_shape.append(1)
             cluster_shape = tuple(cluster_shape)
 
-        shape_np = np.array(shape, dtype=np.int)
+        shape_np = np.array(shape, dtype=int)
         # Softmax on cluster shape gives strong preference to larger dimensions.
         cluster_weights = np.exp(np.array(cluster_shape)) / np.sum(np.exp(cluster_shape))
         shape_fracs = np.array(shape) / np.sum(shape)
@@ -122,7 +122,7 @@ class ArrayApplication(object):
         # Put remainder on largest axis.
         remaining = np.sum(grid_shape_frac - grid_shape)
         grid_shape[np.argmax(shape)] += remaining
-        grid_shape = np.ceil(grid_shape).astype(np.int)
+        grid_shape = np.ceil(grid_shape).astype(int)
 
         # We use ceiling of floating block shape
         # so that resulting grid shape is <= to what we compute above.
@@ -133,6 +133,7 @@ class ArrayApplication(object):
         # Simple way to ensure shape compatibility for basic linear algebra operations.
         block_shape = self.compute_block_shape(shape, dtype)
         final_block_shape = []
+
         for axis in range(len(shape)):
             shape_dim = shape[axis]
             block_shape_dim = block_shape[axis]
@@ -261,7 +262,7 @@ class ArrayApplication(object):
                                                       })
         return rarr
 
-    def read_csv(self, filename, dtype=np.float, delimiter=',', has_header=False, num_workers=None):
+    def read_csv(self, filename, dtype=float, delimiter=',', has_header=False, num_workers=None):
         if num_workers is None:
             num_workers = self.num_cores_total()
         arrays: list = self._filesystem.read_csv(filename, dtype, delimiter, has_header,
@@ -443,14 +444,17 @@ class ArrayApplication(object):
             raise ValueError("X must have 1 or 2 axes.")
         return rarr
 
-    def arange(self, shape, block_shape, step=1, dtype=np.int64) -> BlockArray:
+    def arange(self, start_in, shape, block_shape, step=1, dtype=None) -> BlockArray:
         assert step == 1
+        if dtype is None:
+            dtype = np.__getattribute__(str(np.result_type(start_in, shape[0] + start_in)))
+
         # Generate ranges per block.
         grid = ArrayGrid(shape, block_shape, dtype.__name__)
         rarr = BlockArray(grid, self.system)
         for _, grid_entry in enumerate(grid.get_entry_iterator()):
             syskwargs = {"grid_entry": grid_entry, "grid_shape": grid.grid_shape}
-            start = block_shape[0] * grid_entry[0]
+            start = start_in + block_shape[0] * grid_entry[0]
             entry_shape = grid.get_block_shape(grid_entry)
             stop = start + entry_shape[0]
             rarr.blocks[grid_entry].oid = self.system.arange(start,
@@ -464,9 +468,11 @@ class ArrayApplication(object):
         assert axis == 0
         assert endpoint is True
         assert retstep is False
+
         step_size = (stop - start) / (shape[0] - 1)
-        result = self.arange(shape, block_shape)
+        result = self.arange(0, shape, block_shape)
         result = start + result * step_size
+
         if dtype is not None and dtype != result.dtype:
             result = result.astype(dtype)
         return result
--- a/nums/core/array/utils.py
+++ b/nums/core/array/utils.py
@@ -49,12 +49,12 @@ def is_uint(val, type_test=False):
 
 def is_int(val, type_test=False):
     return is_type(type_test, val,
-                   (int, np.int, np.int8, np.int16, np.int32, np.int64))
+                   (int, np.int8, np.int16, np.int32, np.int64))
 
 
 def is_float(val, type_test=False):
     return is_type(type_test, val,
-                   (float, np.float, np.float16, np.float32, np.float64))
+                   (float, np.float16, np.float32, np.float64))
 
 
 def is_complex(val, type_test=False):
@@ -75,7 +75,7 @@ def get_reduce_output_type(op_name, dtype):
 def shape_from_block_array(arr: np.ndarray):
     grid_shape = arr.shape
     num_axes = len(arr.shape)
-    shape = np.zeros(num_axes, dtype=np.int)
+    shape = np.zeros(num_axes, dtype=int)
     for j in range(num_axes):
         pos = [[0]] * num_axes
         pos[j] = range(grid_shape[j])
--- a/nums/core/settings.py
+++ b/nums/core/settings.py
@@ -69,3 +69,43 @@ np_bop_reduction_set = {
     "nanmin",
     "nansum"
 }
+
+# Fallback on NumPy for these operations.
+# This is achieved by converting the block array to a single block, performing the operation,
+# and converting back to the original block shape.
+doctest_fallback = {
+    'argwhere', 'asscalar', 'clip', 'compress', 'convolve', 'corrcoef', 'cumprod',
+    'cumproduct', 'cumsum', 'diag_indices_from', 'diagflat', 'fix', 'fromiter', 'full',
+    'msort', 'nancumprod', 'nancumsum', 'nanprod', 'partition', 'polysub', 'product',
+    'ravel_multi_index', 'repeat', 'resize', 'roll', 'roots', 'rot90', 'round',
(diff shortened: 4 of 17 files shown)


nums/numpy/__init__.py
```diff
@@ -31,10 +34,39 @@ def _init():
...
+            if hasattr(np, func.__name__) and func.__name__ not in failed_fallback:
+                globals()[name] = _default_to_numpy(func)
+            else:
+                globals()[name] = _not_implemented(func)
+            if hasattr(np, func.__name__):
+                globals()[name].__doc__ = update_doc_string(np.__getattribute__(func.__name__)
+                                                            .__doc__)
 
 
+def update_doc_string(doc_string):
```
elibol, line 50: Is there an appropriate utils module we can move this function to?

Also, if the generated doc string is just for running tests, then can we set the docs using this function in tests and not in production code? Would be great to have updated docs in production code, and I think at this point just providing the numpy doc string as-is is probably ok.

nums/numpy/api.py
```diff
@@ -27,15 +27,29 @@
...
     # From project JAX: https://github.com/google/jax/blob/master/jax/numpy/lax_numpy.py
     def wrapped(*args, **kwargs):
         # pylint: disable=unused-argument
         msg = "NumPy function {} not yet implemented."
-        raise NotImplementedError(msg.format(fun))
+        raise NotImplementedError(msg.format(func))
     return wrapped
 
 
+def _default_to_numpy(func):
```
elibol, line 39: Can you add a TODO here to later convert this function to invoke the numpy op on a worker vs. the driver?

nums/numpy/api.py
```diff
@@ -62,15 +76,14 @@ def wrapped(*args, **kwargs):
...
-int = np.int
 int8 = np.int8
 int16 = np.int16
 int32 = np.int32
 int64 = np.int64
-float = np.float
 float16 = np.float16
 float32 = np.float32
 float64 = np.float64
+float128 = np.float128
```
elibol, line 86: Can you add a space here?

nums/numpy/api.py
```diff
@@ -128,20 +141,22 @@ def array(object, dtype=None, copy=True, order="K", ndmin=0, subok=False) -> Blo
...
+def zeros(shape, dtype=float):
     app = _instance()
     block_shape = app.get_block_shape(shape, dtype)
     return app.zeros(shape=shape, block_shape=block_shape, dtype=dtype)
 
 
-def ones(shape, dtype=np.float):
+def ones(shape, dtype=float):
     app = _instance()
+    if isinstance(shape, int):
```
elibol, line 158: Is integer shape supported by numpy? If so, can we move this somewhere where it will be triggered for other creation functions (e.g. empty and zeros)?

nums/numpy/api.py
```diff
@@ -246,6 +261,10 @@ def atleast_3d(*arys):
 
 
 def arange(start=None, stop=None, step=1, dtype=np.int64) -> BlockArray:
```
elibol, line 263: Can you confirm the semantics for numpy? If we pass floats to this function, does it work or raise an exception?

tests/conftest.py
```diff
@@ -41,6 +41,7 @@ def app_inst(request):
     yield app_inst
     app_inst.system.shutdown()
     ray.shutdown()
+    time.sleep(10)
```
elibol, line 44: Can we set this to 5 (same below)?

elibol (commented):
Looks like a great start! Left some comments. I'm mostly curious about the dtype-related changes. Is it true that `np.int is int` and `np.float is float`?

nums/core/array/application.py
```diff
@@ -71,6 +71,8 @@ def compute_block_shape(self,
             dtype = np.dtype(dtype)
         elif dtype in (bool, np.bool_):
             dtype = np.dtype(np.bool_)
+        elif dtype in (object, np.object_):
```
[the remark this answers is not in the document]
MohamedElgharbawy: In one of the doctests (I can't recall which one, I noticed it randomly), this would error, saying `np.object` was not a valid dtype. NumPy allows it to be used, so I added this line to fix the issue.

nums/api.py
```diff
@@ -68,7 +66,7 @@ def delete(filename: str) -> BlockArray:
         return _instance().delete_fs(filename)
 
 
-def read_csv(filename, dtype=np.float, delimiter=',', has_header=False) -> BlockArray:
+def read_csv(filename, dtype=float, delimiter=',', has_header=False) -> BlockArray:
```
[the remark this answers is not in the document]
MohamedElgharbawy: NumPy kept giving warnings about `np.float`, `np.bool`, and `np.int` being deprecated and to use `float`, `bool`, and `int` instead, so I've been updating these wherever I find them.

nums/core/array/application.py
```diff
@@ -131,8 +133,11 @@ def compute_block_shape(self,
 
     def get_block_shape(self, shape, dtype: np.dtype):
         # Simple way to ensure shape compatibility for basic linear algebra operations.
+        if isinstance(shape, int):
```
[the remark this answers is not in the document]
MohamedElgharbawy: Passing in an int as a shape is valid, but this was erroring since we did not handle that edge case.

nums/core/settings.py
```diff
@@ -69,3 +69,74 @@
...
+    'isposinf', 'isreal', 'isrealobj', 'isscalar', 'ix_', 'kaiser', 'kron', 'lexsort',
+    'maximum_sctype', 'median', 'meshgrid', 'min_scalar_type', 'mintypecode', 'modf', 'moveaxis',
+    'nan_to_num', 'nanargmax', 'nanargmin', 'nanmedian', 'nanpercentile', 'nanquantile', 'nonzero',
+    'obj2sctype', 'packbits', 'pad', 'percentile', 'piecewise', 'place', 'poly', 'polyadd',
+    'polyder', 'polydiv', 'polyfit', 'polyint', 'polymul', 'polyval', 'prod', 'promote_types',
+    'ptp', 'put', 'put_along_axis', 'putmask', 'quantile', 'ravel', 'real', 'real_if_close',
+    'require', 'result_type', 'rollaxis', 'row_stack', 'sctype2char', 'select', 'sinc', 'sort',
+    'stack', 'take', 'take_along_axis', 'trace', 'tril_indices', 'trim_zeros', 'triu_indices',
+    'unique', 'unpackbits', 'unravel_index', 'unwrap', 'vander', 'vdot', 'vsplit', 'vstack', 'who',
+    # Excepted
```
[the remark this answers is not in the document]
MohamedElgharbawy: An exception occured.

nums/core/settings.py
```diff
@@ -69,3 +69,74 @@
     "nanmin",
     "nansum"
 }
+
+# Fallback on NumPy for these operations.
+# This is achieved by converting the block array to a single block, performing the operation,
+# and converting back to the original block shape.
+fallback = {
```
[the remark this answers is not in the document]
MohamedElgharbawy: Yes, I think that might be better too.

nums/numpy/api.py
MohamedElgharbawy: Sure.

nums/numpy/api.py
MohamedElgharbawy: Yep.

tests/conftest.py
MohamedElgharbawy: Yes.

nums/numpy/api.py
MohamedElgharbawy: Yes, it seems like it rounds up to the nearest whole number, so `np.arange(3.1)` outputs `[1, 2, 3, 4]`. I will adjust this implementation to account for this.

nums/numpy/api.py
MohamedElgharbawy: Yes, it is supported. If I'm understanding your request correctly, you'd like to add this to zeros and empty?

nums/numpy/__init__.py
```diff
@@ -31,10 +34,39 @@ def _init():
     # pylint: disable=import-outside-toplevel
     import numpy as np
     from nums.core.systems import utils as system_utils
```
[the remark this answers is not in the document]
MohamedElgharbawy: Will do.

nums/numpy/__init__.py
MohamedElgharbawy: I think this is a good idea, yes.

elibol (commented):
Responded to followup questions.

nums/core/array/utils.py
```diff
@@ -49,12 +49,12 @@ def is_uint(val, type_test=False):
 
 def is_int(val, type_test=False):
     return is_type(type_test, val,
-                   (int, np.int, np.int8, np.int16, np.int32, np.int64))
+                   (int, np.int8, np.int16, np.int32, np.int64))
```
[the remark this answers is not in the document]
MohamedElgharbawy: Correct.

nums/core/array/application.py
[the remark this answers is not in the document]
MohamedElgharbawy: I'm unable to track down which doctest it was, so for now I'll remove this.

nums/core/array/application.py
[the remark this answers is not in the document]
MohamedElgharbawy: Never mind, I believe this error is happening somewhere else. I have removed this.

nums/core/settings.py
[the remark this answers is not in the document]
MohamedElgharbawy: WIll do. FYI, there are no more exceptions after refactoring the fallback approach.

nums/numpy/api.py
MohamedElgharbawy: Sure.

MohamedElgharbawy: Update: Seems like `arange` might have a bug. `nps.arange(3.1, 5.1)` should return `[3.1 4.1]`, but it currently returns `[0, 1]` (without me changing anything). Will look into this.

nums/numpy/api.py
MohamedElgharbawy: Like we discussed offline, I'll add a quick fix for `empty`, `zeros`, and `ones`, and look into a better fix in the future.

nums/numpy/api.py
elibol: Resolved offline.

nums/numpy/api.py
```diff
@@ -27,12 +27,29 @@
...
     def wrapped(*args, **kwargs):
         # pylint: disable=unused-argument
         msg = "NumPy function {} not yet implemented."
-        raise NotImplementedError(msg.format(fun))
+        raise NotImplementedError(msg.format(func))
+
+    return wrapped
+
+
+# TODO(mwe): Convert this to invoke the NumPy op on a worker instead of the driver
```
elibol, line 40: (nit) can you add a space between TODO and your initials `#TODO (mwe): `, and a period at the end of the sentence?

nums/numpy/api.py
```diff
@@ -27,12 +27,29 @@
...
+def _default_to_numpy(func):
+    def wrapped(*args, **kwargs):
```
elibol, line 42: Can we add this warning here: "Operation %s not implemented, falling back to NumPy. If this is too slow or failing, please open an issue on github."

nums/core/array/application.py
```diff
@@ -464,9 +468,13 @@ def linspace(self, start, stop, shape, block_shape, endpoint, retstep, dtype, ax
         assert axis == 0
         assert endpoint is True
         assert retstep is False
-        step_size = (stop - start) / (shape[0] - 1)
-        result = self.arange(shape, block_shape)
-        result = start + result * step_size
+
+        result = self.arange(start, shape, block_shape)
```
elibol, line 472: I think if we call `self.arange(0, shape, block_shape)` here, we can retain the original implementation, which invokes fewer nums ops (2 vs 5). Thoughts?

elibol (commented):
Nice catch with `arange`. Left a few more minor comments. I'm thinking we ought to output a warning whenever a fallback function is invoked. This will help the community (both devs and users) keep track of what we currently support, vs. ops for which we fallback on numpy.

nums/numpy/api.py
MohamedElgharbawy: Sure.

nums/numpy/api.py
MohamedElgharbawy: Okay.

nums/core/array/application.py
MohamedElgharbawy: Let me give it a shot and see if it works.

MohamedElgharbawy: Can confirm it works.

PR #38, opened 2021-04-16
Element-wise broadcasting tests

briancpark:
Adding tests for #33


nums/core/array/application.py
```diff
@@ -715,6 +715,10 @@ def map_bop(self,
         if not (kwargs is None or len(kwargs) == 0):
             raise NotImplementedError("'kwargs' is not yet supported.")
 
+        if not array_utils.broadcastable(arr_1.shape, arr_2.shape,
+                                         arr_1.block_shape, arr_2.block_shape):
+            raise ValueError("operands cannot be broadcasted.")
```
elibol, line 720: Can you capitalize this sentence?

elibol (commented):
Looks good! Left a very minor comment.

tests/numpy/test_arithmetic.py
```diff
@@ -147,6 +147,41 @@ def test_inner_outer(nps_app_inst):
...
 # TODO(hme): Add broadcast tests.
 
+def test_broadcast_block_shape_error(nps_app_inst):
+    from nums import numpy as nps
+    import pytest
+    assert nps_app_inst is not None
+
+    _ops = ['add', 'subtract', 'divide', 'bitwise_and']
+
+    def check_broadcast_block_shape_mismatch_simple_error(_ns_a, _ns_b,
```
MohamedElgharbawy, line 157: Could you shorten this name?

nums/core/array/utils.py
```diff
@@ -118,6 +118,27 @@ def can_broadcast_shapes(a_shape, b_shape):
...
 
 
+def broadcastable(a_shape, b_shape, a_block_shape, b_block_shape):
+    if a_shape == b_shape:
+        return a_block_shape == b_block_shape
+
+    result_shape = broadcast_shape(a_shape, b_shape)
+    if result_shape is None:
+        return False
+    if len(a_shape) == len(b_shape):
```
MohamedElgharbawy, line 128: What is the purpose of this check?

nums/core/array/utils.py
```diff
@@ -118,6 +118,27 @@ def can_broadcast_shapes(a_shape, b_shape):
...
+        return a_block_shape == b_block_shape
+
+    min_bs, max_bs = sorted([a_block_shape, b_block_shape], key=len)
+    for i in range(-1, -len(max_bs) - 1):
+        if -len(min_bs) - 1 < i:
```
MohamedElgharbawy, line 133: What do you think about condensing this into:

```python
if -len(min_bs) - 1 < i and (a_block_shape[i] != b_block_shape[i] or max_bs[i] != 1):
    return False
```

nums/core/array/utils.py
briancpark: I think I discussed this with Melih, as he did most of the PR. I only took over in terms of writing some tests. 

I think it's a case to broadcast block arrays when block shapes matches up? Please correct me if this is wrong @elibol. I can try to make another test specific to that functionality if needed.

elibol: That's right. Put another way, if the two arrays have the same shape length, and the block shapes are the same, then they may be broadcasted.

tests/numpy/test_arithmetic.py
elibol: Agree. Since this is a nested function, we could rename it to `check_value_error`.

elibol (commented):
Left a very minor comment. Can merge once it's addressed.

nums/core/array/utils.py
```diff
@@ -129,6 +129,29 @@ def can_broadcast_shapes(a_shape, b_shape):
...
+def broadcastable(a_shape, b_shape, a_block_shape, b_block_shape):
+    if a_shape == b_shape:
+        return a_block_shape == b_block_shape
+
+    result_shape = broadcast_shape(a_shape, b_shape)
+    if result_shape is None:
+        return False
+    if len(a_shape) == len(b_shape):
+        return a_block_shape == b_block_shape
+    if a_shape == a_block_shape == () or b_shape == b_block_shape == ():
```
nums-project, line 141: This may be an ok fix, but I don't quite understand the original issue. Was it that shapes were `()` and block shapes were `(23,)`, or was the block shape for one of the arrays `()`?

elibol (approved):

PR #39, opened 2021-04-20
Add pylint workflow

bveeramani:
This PR implements a GitHub Action that runs `pylint` when PRs are opened or updated.


setup.py
```diff
@@ -27,7 +27,6 @@
 
 test_requirements = [
     'pytest==6.1.1',
-    'pytest-pylint==0.17.0',
```
nums-project, line 30: Can we re-add this as a requirement for testing? We sometimes want to be able to run linting locally without having to install the required files.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,22 @@
+name: lint
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
```
nums-project, line 5: Could you briefly describe "synchronize?"

bveeramani: Synchronize refers to when you commit changes to an existing PR.

setup.py
bveeramani: Fixed

nums-project:
Is there a way to test that linting is working properly? Is there a way to observe the output of the linting check?

PR #41, opened 2021-04-23
Bump ray version.

elibol:


elibol:
Bumping Ray version to 1.3 introduced a bug in the test infrastructure that triggers a segfault when starting/stopping Ray after performing the operation `rX.T.tensordot(rX, axes=1)` in `test_tensordot_basic`, as well as a related bug in `test_matmul`.

The issue with `test_tensordot_basic` was resolved by performing the same test with different copies of the same data. This suggests an issue when performing the tensordot operation on the same input (which is what happens when we perform tensordot on an input and its transpose).

The issue with `test_matmul` was resolved by generating numpy objects from nums objects instead of nums objects from numpy objects. It's unclear how this change resolved the issue, but both tests perform some variant of `X.T @ X`, which suggests the issue is some side effect of performing a tensordot on the same input data.

Issue #24, opened 2021-02-27
Increasing dimensionality with slicing is not supported


MohamedElgharbawy:
For example, in NumPy:

```
>>> a = np.array([1, 2, 3])
>>> a
array([1, 2, 3])
>>> a[np.newaxis, :]
array([[1, 2, 3]])
```

PR #42, opened 2021-04-23
Implemented stack functions

MohamedElgharbawy:
Implemented:

- `hstack`
- `vstack`
- `dstack`
- `row_stack`
- `column_stack`

Set up the groundwork for `stack`, but implementing [creating a new dimension with indexing](https://github.com/nums-project/nums/issues/24) must be done first.

### DIFF
--- a/nums/core/array/application.py
+++ b/nums/core/array/application.py
@@ -1048,13 +1048,9 @@ class ArrayApplication(object):
         return res
 
     def atleast_1d(self, *arys):
-        # TODO (hme): Refactor this to use check_or_convert_other
         res = []
         for ary in arys:
-            if not isinstance(ary, BlockArray):
-                ary = np.array(ary)
-                block_shape = self.cm.compute_block_shape(ary.shape, ary.dtype)
-                ary = self.array(ary, block_shape)
+            ary = BlockArray.to_block_array(ary, self.cm)
             if ary.ndim == 0:
                 result = ary.reshape(1)
             else:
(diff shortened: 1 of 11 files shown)


elibol (commented):
Looks good, could you merge master into this branch and resolve conflicts?

tests/numpy/test_stack.py
```diff
@@ -91,6 +91,69 @@ def test_atleast_3d(nps_app_inst):
 
     try_multiple_nd(np.atleast_3d, nps.atleast_3d)
 
+
+def test_hstack(nps_app_inst):
```
elibol, line 95: These tests look like they're doing the same thing. Can we consolidate all of these into a single test?

tests/numpy/test_stack.py
```diff
@@ -91,6 +91,69 @@ def test_atleast_3d(nps_app_inst):
...
+    import nums.numpy as nps
+    assert nps_app_inst is not None
+
+    a1 = nps.array((1, 2, 3))
+    a2 = np.array((1, 2, 3))
+    b = np.array((2, 3, 4))
+    assert np.allclose(nps.hstack((a1, b)).get(), np.hstack((a2, b)))
```
elibol, line 102: Can we please test any auto-conversion functionality separately? Here, a nums array for `b` should be used instead of the numpy array.

nums/core/array/application.py
```diff
@@ -1114,3 +1114,31 @@ def atleast_3d(self, *arys):
             return res[0]
         else:
             return res
+
+    def hstack(self, tup):
+        arrs = self.atleast_1d(*tup)
+        if not isinstance(arrs, list):
+            arrs = [arrs]
+        # As a special case, dimension 0 of 1-dimensional arrays is "horizontal"
+        if arrs and arrs[0].ndim == 1:
```
elibol, line 1123: Why the check for `arrs` here? Could `arr` potentially be `None`?

nums/core/array/application.py
```diff
@@ -1114,3 +1114,31 @@ def atleast_3d(self, *arys):
             return res[0]
         else:
             return res
+
+    def hstack(self, tup):
```
elibol, line 1118: For all the stacking functions, could you add the `axis_block_size` parameter and pass it through to `concatenate`? This lets the user (at the `ArrayApplication` level) specify the block size for the axis along which the arrays are being concatenated. This parameter would not be provided at the numpy API level.

elibol (commented):
Looks good! Left some comments.

nums/core/array/application.py
elibol: This is just what's done in numpy -- it may be that the input is empty.

PR #51, opened 2021-04-27
Run pylint on test directory

bveeramani:


tests/core/array/test_basic.py
```diff
@@ -15,7 +15,7 @@
 
 
 # pylint: disable=wrong-import-order
-import common
+import common  # pylint: disable=import-error
```
elibol, line 18: Does this directive only apply to this line?

elibol: @bveeramani bumping this.

bveeramani: Yes, it should only apply to that line

bveeramani: I'm like 95% sure but I can double-check.

PR #61, opened 2021-05-28
New Diagonal Implementation for non-square matrices

Priyansdesai:
As per the algorithm we created. 

Reference: https://gist.github.com/Priyansdesai/0607b826f669fd7d290377eda85231e3


elibol (commented):
Looking good. Left some comments.

nums/core/array/application.py
```diff
@@ -332,28 +335,67 @@ def diag(self, X: BlockArray) -> BlockArray:
...
+            block, diagonal, block_no, current_block_length = X.blocks[(0, 0)], OrderedDict(), 1, 0
+            while count <= total_elements:
+                if element_i > block.shape[0] - 1:
+                    block_i = block_i + 1
+                    element_i = 0
+                if element_j > block.shape[1] - 1:
+                    block_j = block_j + 1
+                    element_j = 0
+                block = X.blocks[(block_i, block_j)]
+                while True:
```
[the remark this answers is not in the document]
Priyansdesai: We cannot. Because even if we aren't doing multiple input blocks => one output block, we need the inner loop for the elements we need from a particular block. I think we already discussed about the importance of the inner loop in our last meeting. Probably the additions related to the multiple input blocks stuff makes it look like if it isn't needed.

nums/core/array/application.py
```diff
@@ -332,28 +335,67 @@ def diag(self, X: BlockArray) -> BlockArray:
...
-            for grid_entry in X.grid.get_entry_iterator():
-                out_grid_entry = grid_entry[:1]
-                out_grid_shape = grid.grid_shape[:1]
-                syskwargs = {"grid_entry": out_grid_entry, "grid_shape": out_grid_shape}
-                if np.all(np.diff(grid_entry) == 0):
-                    # This is a diagonal block.
-                    rarr.blocks[out_grid_entry].oid = self.cm.diag(X.blocks[grid_entry].oid,
-                                                                   syskwargs=syskwargs)
+            block_i, block_j, element_i, element_j = 0, 0, 0, 0
+            count, total_elements = 1, min(shape)
```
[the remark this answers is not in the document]
Priyansdesai: I think that's should have been just shape[0]

nums/core/array/application.py
```diff
@@ -332,28 +335,67 @@ def diag(self, X: BlockArray) -> BlockArray:
...
                                                                     grid_entry,
                                                                     grid_meta,
                                                                     syskwargs=syskwargs)
         elif len(X.shape) == 2:
-            assert X.shape[0] == X.shape[1], "X must be a square array."
-            assert X.block_shape[0] == X.block_shape[1], "block_shape must be square."
-            shape = X.shape[0],
-            block_shape = X.block_shape[0],
+            shape = min(X.shape),
+            block_shape = min(X.block_shape),
```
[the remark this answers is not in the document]
Priyansdesai: Will rename them. Also, this entire code comes from the earlier code we had for diagonal.

nums/core/array/application.py
```diff
@@ -332,28 +335,67 @@ def diag(self, X: BlockArray) -> BlockArray:
...
-                out_grid_entry = grid_entry[:1]
-                out_grid_shape = grid.grid_shape[:1]
-                syskwargs = {"grid_entry": out_grid_entry, "grid_shape": out_grid_shape}
-                if np.all(np.diff(grid_entry) == 0):
-                    # This is a diagonal block.
-                    rarr.blocks[out_grid_entry].oid = self.cm.diag(X.blocks[grid_entry].oid,
-                                                                   syskwargs=syskwargs)
+            block_i, block_j, element_i, element_j = 0, 0, 0, 0
+            count, total_elements = 1, min(shape)
+            block, diagonal, block_no, current_block_length = X.blocks[(0, 0)], OrderedDict(), 1, 0
```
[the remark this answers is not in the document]
Priyansdesai: Yes. I have added the explanation for all of these additions below. Will also add comments in the file.

nums/core/array/application.py
```diff
@@ -332,28 +335,67 @@ def diag(self, X: BlockArray) -> BlockArray:
...
+                    element_i = 0
+                if element_j > block.shape[1] - 1:
+                    block_j = block_j + 1
+                    element_j = 0
+                block = X.blocks[(block_i, block_j)]
+                while True:
+                    block_rows, block_cols = block.shape[0], block.shape[1]
+                    if element_i > block_rows - 1 or element_j > block_cols - 1:
+                        break
+                    if current_block_length == block_shape[0]:
```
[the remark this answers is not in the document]
Priyansdesai: Check my last comment.

nums/core/array/application.py
```diff
@@ -332,28 +335,67 @@ def diag(self, X: BlockArray) -> BlockArray:
...
+                        block_no += 1
+                        current_block_length = 1
+                    else:
+                        current_block_length += 1
+                    block_dict = diagonal.get(block_no, OrderedDict())
+                    elements = block_dict.get((block_i, block_j), [])
+                    elements.append((element_i, element_j))
+                    block_dict[(block_i, block_j)] = elements
+                    diagonal[block_no] = block_dict
+                    count, element_i, element_j = count + 1, element_i + 1, element_j + 1
```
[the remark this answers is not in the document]
Priyansdesai: Sure we can do that. In fact, this could go into utils itself, if we pass in the right parameters, this doesn't need anything but certain sizes.

Priyansdesai:
Hey Melih!

I am going to explain the whole new code addition in various parts of our earlier code below. 

**1. First, let's consider why at all we needed these changes, i.e. the case where we need multiple input blocks to compute an output block?** 

Let's consider an identity matrix of size 4 by 4 and it is made of blocks of size (2, 3). So, our diagonal passes through 3 blocks (0, 0), (1, 0) and (1, 1). But, our output block_shape can only be the [min(2, 3), ] since every block will at most have 2 elements and overall shape of the grid can only be [min(4, 4), ], because whichever is the minimizing parameter, that would be the no. of elements we will have in our diagonal. So, in this case, that would be we will have 2 blocks of each of size [2, 1]. But, if you see earlier, there are 3 blocks which have our elements: (0, 0), (1, 0) and (1, 1). So, here we would have to combine the last 2 blocks, since the first block would result in output block of size (2, 1). 

**2. Here we will consider how we achieve this in the code.**

First, I will explain the overall approach of how I have done this and then will explain the importance of each variable. So, the approach I am taking is maintaining this variable called `current_block_length` which keeps track of the no. of elements in a particular output block. Whenever this variable becomes greater than `block_shape[0]`, we make a new block and reset `current_block_length`. Now what does, making a block here mean. To understand this, let's see how I store the information for a particular output block. For every output block, I have a hashmap/dictionary with the keys being the block indices from the main Array Grid and the values being list of elements that I am taking from that block. This ensures a block can be split between two output blocks, hence every input block can be used for at most 2 output blocks. 

-

nums/core/compute/numpy_compute.py
```diff
@@ -169,8 +169,12 @@ def update_block_along_axis(self, dst_arr, src_arr, index_pairs, axis):
             result[tuple(dst_sel)] = src_arr[tuple(src_sel)]
         return result
 
-    def diag(self, arr):
-        return np.diag(arr)
+    def diag(self, arr, elements):
+        if elements == []:
+            return np.diag(arr)
+        else:
+            rarr = list(map(lambda index: arr[index[0]][index[1]], elements))
```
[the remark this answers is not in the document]
Priyansdesai: I think it is quite doable.

nums-project:
> Hey Melih!
> 
> I am going to explain the whole new code addition in various parts of our earlier code below.
> 
> **1. First, let's consider why at all we needed these changes, i.e. the case where we need multiple input blocks to compute an output block?**
> 
> Let's consider an identity matrix of size 4 by 4 and it is made of blocks of size (2, 3). So, our diagonal passes through 3 blocks (0, 0), (1, 0) and (1, 1). But, our output block_shape can only be the [min(2, 3), ] since every block will at most have 2 elements and overall shape of the grid can only be [min(4, 4), ], because whichever is the minimizing parameter, that would be the no. of elements we will have in our diagonal. So, in this case, that would be we will have 2 blocks of each of size [2, 1]. But, if you see earlier, there are 3 blocks which have our elements: (0, 0), (1, 0) and (1, 1). So, here we would have to combine the last 2 blocks, since the first block would result in output block of size (2, 1).

Okay, in this case we ought to forego creating the output block and create a separate `BlockArray` for each piece of the diagonal extracted from each block. In the example you provided, we would have 3 block arrays: One with 2 elements, and two with 1 element. We can then use the concatenate method to stitch them together. This will effectively do what you're doing manually. The advantage to using concatenate is maintainability and readability.

> **2. Here we will consider how we achieve this in the code.**
> 
> First, I will explain the overall approach of how I have done this and then will explain the importance of each variable. So, the approach I am taking is maintaining this variable called `current_block_length` which keeps track of the no. of elements in a particular output block. Whenever this variable becomes greater than `block_shape[0]`, we make a new block and reset `current_block_length`. Now what does, making a block here mean. To understand this, let's see how I store the information for a particular output block. For every output block, I have a hashmap/dictionary with the keys being the block indices from the main Array Grid and the values being list of elements that I am taking from that block. This ensures a block can be split between two output blocks, hence every input block can be used for at most 2 output blocks.
> 
> * diagonal: a dictionary where the key is the output block no and the value is the dictionary containing the information below.
> * block_dict: hashmap/dictionary with the keys being the block indices from the main Array Grid and the values being list of elements that I am taking from that block.
> * block_no: in which output block are we currently adding information.
> * current_block_length: how many elements have we added so far to the current block
> 
> An example of how diagonal looks like:
> `diagonal = {1: {(0, 1): [(0, 0), (1, 1)]}, 2: {(1, 0): [(0, 2)], (1, 1): [(0, 1)]}`
> 
> Now, the last block of code:
> Here what we need to do is for every output block, we get the diagonals from the diag function in numpy_compute.py and store the oids of all output arrays. Then we create slices of the input array and the output array using the find_slices function to pass into the create_block function.
> 
> Edit as of 12:30PM PST on 2nd June 2021:
> I have edited the code. And here is the gist of the new util functions.
> 
> Gist: https://gist.github.com/Priyansdesai/5df73fb0917c996457e72101d8cfd983

Could you summarize the last two parts as brief comments in the code?

nums/core/array/utils.py
```diff
@@ -26,6 +27,60 @@
...
+        src_params.append((slice_obj_src, False))
+        dst_params.append((slice_obj_dst, False))
+        count += length
+    return src_params, dst_params, count
+
+def find_output_blocks(X_blocks, out_shape, out_block_shape):
+    block_i, block_j, element_i, element_j = 0, 0, 0, 0
+    count, total_elements = 0, out_shape[0]
+    block, diagonal = X_blocks[(0, 0)], OrderedDict()
+    block_no, curr_block_length, blocks = 1, 0, []
```
nums-project, line 46: This is still difficult to read. Can you write these out on separate lines, and describe their roles briefly in the code? I understand what they do. We want to make it as easy as possible for future contributors to understand what's going on here.

nums-project: Also, change the name of blocks to `diag_meta` or something more interpretable. When I see a list named `blocks`, I think this is going to be a list of type `Block`.

nums/numpy/api.py
```diff
@@ -93,7 +93,7 @@ def wrapped(*args, **kwargs):
 float16 = np.float16
 float32 = np.float32
 float64 = np.float64
-float128 = np.float128
+#float128 = np.float128
```
nums-project, line 96: Why is this commented out?

nums/core/utils.py
```diff
@@ -11,4 +11,4 @@
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
-# limitations under the License.
+# limitations under the License.
```
nums-project, line 14: Why was this file modified?

nums/core/array/application.py
```diff
@@ -332,28 +334,45 @@ def diag(self, X: BlockArray) -> BlockArray:
...
+            
+            grid_i, grid_j = 0, 0
+            for block_no in diagonal:
+                syskwargs = {"grid_entry": (grid_i,) ,"grid_shape": out_grid_shape}
+                blocks = diagonal[block_no]
+                if len(blocks) == 1:
+                    rarr.blocks[(grid_i, )].oid = all_result_oids[list(blocks.keys())[0]]
+                else:
+                    all_oids = [all_result_oids[key] for key in blocks]
+                    src_params, dst_params,total = array_utils.find_slices(blocks)
```
nums-project, line 367: As mentioned in response to your description of your code, I think we can remove `find_slices` and the use of `create_block` in general and just use `concatenate` instead. Do you see any issue with this? This would improve the readability and maintainability of the code.

nums-project (commented):
Looking good! Left a few more comments.

nums/core/array/utils.py
Priyansdesai: Will do!

nums/core/utils.py
Priyansdesai: Actually, remember that I had added the find_slices function incorrectly over here..so I deleted that.

nums/numpy/api.py
Priyansdesai: Actually, this was causing an error on my system. I had posted about this on the Slack and Mohamed said that just comment it out for now as he had added that. I will add it back.

nums/core/array/application.py
Priyansdesai: Done. Actually I tried doing using `concetenate` but at that time I didn't do correctly and moved to this solution. But using `concetenate` is definitely better and I have implemented that.

Priyansdesai:
Pushes new changes. @elibol @nums-project

nums/core/array/application.py
```diff
@@ -332,28 +331,37 @@ def diag(self, X: BlockArray) -> BlockArray:
...
-                    # This is a diagonal block.
-                    rarr.blocks[out_grid_entry].oid = self.cm.diag(X.blocks[grid_entry].oid,
-                                                                   syskwargs=syskwargs)
+            out_shape = min(X.shape),
+            out_block_shape = min(X.block_shape),
+            # Obtain the block indices which contain the diagonal of the matrix.
+            diag_meta = array_utils.find_output_blocks(X.blocks, out_shape[0])
+
+            all_result_blocks = []
+            out_grid_shape, count = (len(diag_meta),), 0
```
nums-project, line 348: Please separate these into two lines. In general, define variables on separate lines. It took me multiple scans to identify where `count` was being defined.

nums/core/array/application.py
```diff
@@ -332,28 +331,37 @@ def diag(self, X: BlockArray) -> BlockArray:
...
+            # Obtain the diagonals.
+            for block_indices, offset, total_elements in diag_meta:
+                syskwargs = {"grid_entry": (count,) ,"grid_shape": out_grid_shape}
+                result_block_shape = total_elements,
+                block_grid = ArrayGrid(result_block_shape, result_block_shape,
+                                        X.blocks[block_indices].dtype.__name__)
+                block = BlockArray(block_grid, self.cm)
```
nums-project, line 355: Can you rename `block` to `ba` or `block_array`?

nums/core/array/application.py
```diff
@@ -332,28 +331,37 @@ def diag(self, X: BlockArray) -> BlockArray:
...
+                block.blocks[0].oid = self.cm.diag(X.blocks[block_indices].oid,
+                                                 offset, syskwargs=syskwargs)
+                all_result_blocks.append(block)
```
nums-project, line 358: Can you rename `all_result_blocks` to `block_arrays` or something similar?

nums/core/array/application.py
```diff
@@ -332,28 +331,37 @@ def diag(self, X: BlockArray) -> BlockArray:
...
+                count += 1
+            if len(all_result_blocks) > 1:
+                # If there are multiple blocks, concatenate them.
+                return self.concatenate(all_result_blocks, axis=0,
+                                        axis_block_size=out_block_shape[0])
+            return block
```
nums-project, line 364: Instead of returning `block`, can you return `all_result_blocks[0]`? While it's true in Python, in general it's not good practice to assume variables defined in loops exist outside of the scope of the loop.

nums/core/array/utils.py
```diff
@@ -26,6 +26,37 @@
 # pylint: disable = no-member, trailing-whitespace
 
 
+def find_output_blocks(X_blocks, total_elements):
```
nums-project, line 29: Can we rename this to `find_diag_output_blocks`?

nums-project (commented):
This is coming together very nicely! Great job! I left just a few comments.

What happened to the tests we had for this PR?

nums/core/array/application.py
Priyansdesai: Done. Sorry about that.

nums/core/array/application.py
Priyansdesai: Yes absolutely. Done.

nums/core/array/application.py
Priyansdesai: Yes I renamed it to `output_block_arrays`.

nums/core/array/application.py
Priyansdesai: Yeah I was thinking about the same. But, I have changed it now.

nums/core/array/utils.py
Priyansdesai: Yes done.

nums/core/array/utils.py
```diff
@@ -26,6 +26,46 @@
 # pylint: disable = no-member, trailing-whitespace
 
 
+def find_diag_output_blocks(X_blocks, total_elements):
```
elibol, line 29: Can you nest this function inside of `diag` in `ArrayApplication`? I did not realize it's operating on the blocks of X directly, and `utils` thus far does not operate on blocks directly. Also, let's change the signature of this function to `X: BlockArray, total_elements: int`. Within the body of the code, just use `X.blocks` instead of `X_blocks`.

nums/core/array/application.py
```diff
@@ -364,28 +363,80 @@ def diag(self, X: BlockArray) -> BlockArray:
...
-                out_grid_shape = grid.grid_shape[:1]
-                syskwargs = {"grid_entry": out_grid_entry, "grid_shape": out_grid_shape}
-                if np.all(np.diff(grid_entry) == 0):
-                    # This is a diagonal block.
-                    rarr.blocks[out_grid_entry].oid = self.cm.diag(
-                        X.blocks[grid_entry].oid, syskwargs=syskwargs
+            out_shape = (min(X.shape),)
+            out_block_shape = (min(X.block_shape),)
+            # Obtain the block indices which contain the diagonal of the matrix.
+            def find_diag_output_blocks(X: BlockArray, total_elements: int):
```
elibol, line 376: Let's move this to line 355, right before the `if` condition.

PR #65, opened 2021-06-07
Bump Ray to 1.4.0

vinamrabenara:


nums-project:
Could we look into reversing the changes in test_bop.py from PR #41? We temporarily changed some tests to allow nums tests on Ray 1.3 to run on travis.

vinamrabenara:
> Could we look into reversing the changes in test_bop.py from PR #41? We temporarily changed some tests to allow nums tests on Ray 1.3 to run on travis.

done

PR #72, opened 2021-06-29
Adds isnan.

nums-project:


nums/core/array/application.py
```diff
@@ -234,7 +234,14 @@ def loadtxt(
     def scalar(self, value):
         return BlockArray.from_scalar(value, self.cm)
 
-    def array(self, array: np.ndarray, block_shape: tuple = None):
+    def array(self, array: Union[np.ndarray, List[float]], block_shape: tuple = None):
```
vinamrabenara, line 237: why is this now an union of array type and list of floats?

elibol: This is to add support for list and ndarray types. It's not needed for this PR, but I thought we already supported this. It makes the tests easier to write and is a common usage of `np.array`.

PR #73, opened 2021-06-29
Adds LBFGS.

elibol:


nums/models/glms.py
```diff
@@ -350,10 +382,6 @@ def hessian(self, X: BlockArray, y: BlockArray, mu: BlockArray = None):
 PoissonRegressor = PoissonRegression
 
 
-def line_search():
```
vinamrabenara, line 353: is line search now part of nums/models/lbfgs.py (BackTrackingLineSearch)?

elibol: Yes.

nums/models/glms.py
```diff
@@ -462,8 +490,21 @@ def irls(
...
+def lbfgs(
+    app: ArrayApplication,
+    model: GLM,
+    beta,
+    X: BlockArray,
+    y: BlockArray,
+    tol: BlockArray,
+    max_iter: int,
+):
+    # TODO (hme): Enable a way to provide memory length and line search parameters.
```
vinamrabenara, line 503: Where exactly do we want to provide line search params. What kind of params would these be. would they be kind of line search, like backtracking, etc and max iters and other line search params like alpha etc..

tests/models/test_lbfgs.py
```diff
@@ -0,0 +1,57 @@
...
+
+import numpy as np
+from nums.models.glms import LogisticRegression
+from nums.core.array.application import ArrayApplication
+
+
+def sample_set(app: ArrayApplication):
+    shape = (500, 10)
+    block_shape = (100, 10)
+    rs = app.random_state(1337)
```
vinamrabenara, line 25: is 1337 a seed

elibol: Yes.

nums/models/glms.py
elibol: Yes, I think the best solution here is to allow a way to provide those params via the GLM class constructor `__init__`. For backtracking, you can see a number of parameters hardcoded in the LBFGS class, including initial alpha value, how much it is reduced per iteration, and the slope param.

PR #74, opened 2021-06-30
Add Autoscaler

vinamrabenara:
* Adds a top level folder named autoscaler that contains
   * An configurable yaml file to launch ray clusters on aws
   * A Readme for basic instructions on how to use it
   * An example

Currently only for AWS


autoscaler/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+* Modify the keys in ```provider``` field for your aws specific configurations regarding regions and availability zones. 
+
+* In the ```available_node_types``` field, edit the ```node_config``` field for ```nums.head.default```. This will configure the head node. You can choose the ec2 instance type, disk and AMI for head node here. Similarly, for the ```nums.worker.default``` field, edit the ```min_workers``` key to set the number of NumS workers. Edit the node configurations here as well under ```node_config``` field. (Make sure to use the correct AMI as per your region and availability zones).
+
+* Then modify the ```file_mounts``` field to indicate any directories or files to copy from the local machine to the every node on the cluster. Leave commented otherwise.
+
+## Running
+
+To launch a NumS cluster then run
```
elibol, line 18: "To launch a NumS cluster, run"

For these instructions, could you briefly explain what happens after running the commands? Do they have a Ray cluster with nums? Could we tie this to additional instructions explaining how someone might run a very simple nums program on the cluster?

vinamrabenara: yes, sounds good. will add an example.

autoscaler/autoscaler-aws.yaml
```diff
@@ -0,0 +1,169 @@
...
+    - /usr/bin/python3 -m pip install --upgrade pip
+    - pip3 install -U protobuf
+    - pip3 install psutil
+    - conda deactivate;
+        conda env remove --name nums -y;
+        conda create --name nums python=3.7 -y;
+    - echo "conda activate nums" >> /home/ubuntu/.bashrc
+    - val=$(egrep 'MemTotal' /proc/meminfo | egrep -o '[0-9]{1,99}');
+      unit=K;
+      sudo mount -o remount,size=${val}${unit} /dev/shm;
```
vinamrabenara, line 144: i have added this here

elibol (approved):

PR #76, opened 2021-07-02
Fixes regularization of binary logistic regression.

elibol:


nums/models/glms.py
```diff
@@ -286,7 +286,8 @@ def hessian(self, X: BlockArray, y: BlockArray, mu: BlockArray = None):
         if self._penalty is None:
             return X.T @ (s * X)
```
vinamrabenara, line 287: On page 7 [here](https://www.cs.ubc.ca/~nando/540-2013/lectures/l11.pdf) , hessian is X.T @ (diag(s) * S). Am i missing something?

nums/models/glms.py
```diff
@@ -286,7 +286,8 @@ def hessian(self, X: BlockArray, y: BlockArray, mu: BlockArray = None):
...
         else:
-            return X.T @ (s * X) + self._lambda_vec
+            # TODO (hme): Construct diag of _lambda_vec once.
+            return X.T @ (s * X) + self._app.diag(self._lambda_vec)
```
vinamrabenara, line 290: I am not sure why we take ```diag``` of ```self._lambda_vec```. [Here](https://www.cs.ubc.ca/~nando/540-2013/lectures/l11.pdf) on page 7, diag is instead of ```s```

PR #94, opened 2021-07-09
Implements median and top_k

dlzou:
Both use `quickselect` as a subprocedure. Both currently only support 1D arrays.

`median` has been benchmarked on an EC2 cluster and shown to scale linearly with array size.

`top_k` is technically not part of the NumPy API, but instead a function found in PyTorch and TensorFlow. The closest NumPy equivalent is `np.partition(a, k - 1)[:k]`.


nums/numpy/api.py
```diff
@@ -606,6 +618,47 @@ def average(
...
+        a: A BlockArray.
+        k: Number of top elements to return.
+        largest: Whether to return largest or smallest elements.
+
+    Returns:
+        A tuple containing two BlockArrays, (`values`, `indices`).
+        values: Values of the top k elements, unsorted.
+        indices: Indices of the top k elements, ordered by their corresponding values.
+    """
+    if sorted:
```
elibol, line 657: If you have an idea on how to support this op with the `sorted` argument, you could note it here as a comment.

elibol (changes requested):
Looks great! I left one comment that needs to be addressed. I think we can merge once it's completed.

elibol (approved):

Issue #60, opened 2021-05-24
bitwise_and Extra Error Handling


briancpark:
Currently, bitwise_and throws extra errors with NumS with improper inputs. (Floating points should not be a valid input to `bitwise_and`)

```python
from nums import numpy as nps
a = nps.random.randn(20, 20)
b = nps.random.randn(20, 20)
nps.bitwise_and(a, b)
```
```
Traceback (most recent call last):
  File "/Users/brianpark/Desktop/NumS/fork/nums/core/array/application.py", line 649, in map_bop
    result_blocks: np.ndarray = ufunc(arr_1.blocks, arr_2.blocks)
TypeError: unsupported operand type(s) for &: 'Block' and 'Block'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/brianpark/Desktop/NumS/fork/nums/core/array/utils.py", line 40, in get_bop_output_type
    dtype = np.__getattribute__(op_name)(a, b).dtype
TypeError: ufunc 'bitwise_and' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/brianpark/Desktop/NumS/fork/nums/numpy/api.py", line 964, in bitwise_and
    kwargs=numpy_utils.ufunc_kwargs(kwargs))
  File "/Users/brianpark/Desktop/NumS/fork/nums/core/array/application.py", line 654, in map_bop
    rarr = self._broadcast_bop(op_name, arr_1, arr_2)
  File "/Users/brianpark/Desktop/NumS/fork/nums/core/array/application.py", line 681, in _broadcast_bop
    arr_2.dtype)
  File "/Users/brianpark/Desktop/NumS/fork/nums/core/array/utils.py", line 43, in get_bop_output_type
    dtype = scipy.special.__getattribute__(op_name)(a, b).dtype
AttributeError: module 'scipy.special' has no attribute 'bitwise_and'
```

While NumPy doesn't have the extra `AttributeError`:
```python
import numpy as np
a = np.random.randn(20, 20)
b = np.random.randn(20, 20)
np.bitwise_and(a, b)
```

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: ufunc 'bitwise_and' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

PR #120, opened 2021-07-22
Correcting binary and unary operation error checks

briancpark:
Fixes Issue #60. Properly returns `TypeError` instead of `AttributeError` with a cleaner error log. Tests added make sure they are the same output as NumPy error messages.


tests/numpy/test_arithmetic.py
```diff
@@ -78,6 +78,50 @@ def check_bop(_name, _np_a, _np_b):
...
+
+        _np_a = np.random.randn(2, 2)
+        _np_b = np.random.randn(2, 2)
+
+        _ns_a = nps.array(_np_a)
+        _ns_b = nps.array(_np_b)
+        message = ""
+
+        try:
+            if _name in uops_names:
```
elibol, line 99: Why do we need to perform this test? If the names of the operations are different, can we pass the numpy names as well as the nums names?

tests/numpy/test_arithmetic.py
```diff
@@ -78,6 +78,50 @@ def check_bop(_name, _np_a, _np_b):
                 check_bop(name, np_a, np_b)
 
 
+def test_bitwise_error(nps_app_inst):
+    assert nps_app_inst is not None
+
+    uops, bops = numpy_utils.ufunc_op_signatures()
+    uops_names, bops_names = [_op[0] for _op in uops], [_op[0] for _op in bops]
```
elibol, line 85: If possible, let's try to remove the need for this.

elibol (commented):
Left some comments.

PR #136, opened 2021-08-19
Remove mypy job

bveeramani:
Temporarily removing the mypy job. I'll re-add the job once we've fixed the type annotations.


elibol (approved):
LGTM

PR #139, opened 2021-08-26
Bump up Ray

briancpark:
Recent versions of Ray should now be supported. (Ray 1.0.0 - 1.6.0)

### DIFF
--- a/nums/core/storage/utils.py
+++ b/nums/core/storage/utils.py
@@ -65,7 +65,7 @@ class Batch(object):
 def reverse_readline(filename, buf_size=8192):
     # https://stackoverflow.com/questions/2301789/how-to-read-a-file-in-reverse-order
     """A generator that returns the lines of a file in reverse order"""
-    with open(filename, encoding="UTF-8") as fh:
+    with open(filename, encoding="utf-8") as fh:
         segment = None
         offset = 0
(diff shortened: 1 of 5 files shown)


briancpark:
@elibol @bveeramani Regardless of whether it was pylint being updated, I looked more into it and seems like we're only failing the code W1514. According to this [issue](https://github.com/PyCQA/pylint/issues/4888), encoding can be different between different OS and explicitly setting encoding is recommended. Or we can ignore the error through pylint.

I pushed a commit that quickly fixes the encoding to UTF-8. Do you think this is okay or should we just ignore the error in pylint?

elibol:
> @elibol @bveeramani Regardless of whether it was pylint being updated, I looked more into it and seems like we're only failing the code W1514. According to this [issue](https://github.com/PyCQA/pylint/issues/4888), encoding can be different between different OS and explicitly setting encoding is recommended. Or we can ignore the error through pylint.
> 
> I pushed a commit that fixes the encoding to UTF-8. Do you think this is okay or should we just ignore the error in pylint?

I think it makes sense to be explicit. Still curious why we're getting linting issues with existing code. @bveeramani, any thoughts on this?

elibol:
LGTM, but we'll need to run performance tests before merging this branch.

elibol (approved):

PR #144, opened 2021-09-01
Pin `pylint` version to 2.10.2

### LINKED ISSUE #134: Replace `pytest-pylint` dependency with `pylint`
We currently list `pytest-pylint` as a dependency.

https://github.com/nums-project/nums/blob/7dfb540592e3efcf52fe50da33158b9f4e3360a7/setup.py#L30

However, we use `pylint` as a standalone tool rather than as part of `pytest`.

https://github.com/nums-project/nums/blob/7dfb540592e3efcf52fe50da33158b9f4e3360a7/.github/workflows/lint.yml#L8-L23

We should therefore replace the `pytest-pylint` dependency with `pylint`.

bveeramani:
Fixes #134


elibol (approved):
LGTM

PR #146, opened 2021-09-08
Implementation of Upper-Triangular matrix inverse and matrix inverse using Cholesky

aatifjiwani:
For our CS267 final project in Spring 2021, we implemented the algorithm from [Bientenisi et al.](https://www.cs.utexas.edu/users/flame/pubs/siam_spd.pdf) in the NumS framework where we use Ray to take the inverse of an upper-triangular matrix and matrix inversion using blocked Cholesky. 

This PR includes:
- An additional numpy API `sum_reduce` which is required for Upper-Triangular matrix inversion
- Implementation in `linalg` module for Upper-Tri matrix inversion
- Implementation in `linalg` of blocked Cholesky
- Using blocked Cholesky implementation, we include an implementation of Cholesky matrix inversion which also uses the upper-triangular matrix inversion


nums/core/linalg.py
```diff
@@ -256,6 +256,136 @@ def inv(app: ArrayApplication, X: BlockArray):
...
+
+                # Calculate -R00 = 0 - R00
+                neg_R00_oid = app.cm.bop("subtract", Z_oid, R00_oid,
+                                                False, False, axes=1, syskwargs={
+                                                    "grid_entry": (row_block, col_block),
+                                                    "grid_shape": grid_shape
+                                                })
+                # Calculate -R00 @ R01
+                sub_oids.append(app.cm.bop("tensordot", neg_R00_oid, R01_oids[col_block],
+                                    False, False, axes=1, syskwargs={
```
aatifjiwani, line 339: FYI @elibol @dlzou Thanks for the heads-up on the API change! I tested the test locally and it passes, so hopefully the build will pass as well.

aatifjiwani: Linting errors should be fixed now as well

aatifjiwani: ugh... linting is a pain

vinamrabenara:
@aatifjiwani did you try running ```black nums/core/linalg.py``` to fix the black error?

aatifjiwani:
@vinamrabenara Done!

tests/core/test_linalg.py
```diff
@@ -215,7 +228,8 @@ def test_rr(app_inst: ArrayApplication):
     app_inst = conftest.get_app("ray")
     # test_inv_assumptions(app_inst)
     # test_inv(app_inst)
-    test_qr(app_inst)
+    test_inv_uppertri(app_inst)
+    # test_qr(app_inst)
     # test_svd(app_inst)
```
vinamrabenara, line 233: Did you forget to uncomment the other tests?

vinamrabenara, line 233: I was wondering why these other tests are commented out. @elibol

aatifjiwani: I believe it is commented out for testing locally. `pytest` should run all the methods with `test_` in the build checks.

nums/core/linalg.py
```diff
@@ -256,6 +256,160 @@ def inv(app: ArrayApplication, X: BlockArray):
     return result
 
 
+def inv_uppertri(app: ArrayApplication, X: BlockArray):
+    # Inversion of an Upper Triangular Matrix
+    # Use the method described in https://www.cs.utexas.edu/users/flame/pubs/siam_spd.pdf
+    assert X.shape[0] == X.shape[1], "This function only accepts square matrices"
```
vinamrabenara, line 262:
```suggestion
    assert len(X.shape) == 2
    assert X.shape[0] == X.shape[1], "This function only accepts square matrices"
```

nums/core/linalg.py
```diff
@@ -279,6 +433,100 @@ def cholesky(app: ArrayApplication, X: BlockArray):
     return result
 
 
+def cholesky_blocked(app: ArrayApplication, X: BlockArray):
```
vinamrabenara, line 436:
```suggestion
def _cholesky_blocked(app: ArrayApplication, X: BlockArray):
```

nums/core/linalg.py
```diff
@@ -256,6 +256,160 @@ def inv(app: ArrayApplication, X: BlockArray):
...
+    # By the time we finish, R = R_inv
+    return R
+
+
+def inv_cholesky(app: ArrayApplication, X: BlockArray):
+    # Matrix Inversion for X where X is a square positive definite matrix
+    # Particularly useful for least-squares regression on a large data corpus
+
+    # Step 1: Calculate the U using Cholesky, where X = U^TU
+    U = app.cholesky_blocked(X)
```
vinamrabenara, line 404:
```suggestion
    U = app._cholesky_blocked(X)
```

tests/core/test_linalg.py
```diff
@@ -125,6 +125,19 @@ def test_inv(app_inst: ArrayApplication):
...
+
+    R = app_inst.array(R_np, block_shape=block_shape)
+
+    R_inv_naive = linalg.inv(app_inst, R).get()
+    R_inv = linalg.inv_uppertri(app_inst, R).get()
+
+    assert np.allclose(R_inv_naive, R_inv)
+
+
 def test_qr(app_inst: ArrayApplication):
```
vinamrabenara, line 141: Can you add tests for ```inv_cholesky```

tests/core/test_linalg.py
```diff
@@ -125,6 +125,19 @@ def test_inv(app_inst: ArrayApplication):
         assert np.allclose(np.linalg.cholesky(mat.get()), L, rtol=1e-4, atol=1e-4)
 
 
+def test_inv_uppertri(app_inst: ArrayApplication):
+    block_shape, shape = (3, 3), (20, 20)
```
vinamrabenara, line 129: i was wondering if it would be a good idea to test with some random ```block_shapes``` and ```shapes``` and then loop over those using ```itertools.product```

tests/core/test_linalg.py
```diff
@@ -125,6 +125,19 @@ def test_inv(app_inst: ArrayApplication):
...
+    R_np = np.arange(1, shape[0] ** 2 + 1, 1).reshape(*shape).astype(float)
+    R_np = np.triu(R_np)
+
+    R = app_inst.array(R_np, block_shape=block_shape)
+
+    R_inv_naive = linalg.inv(app_inst, R).get()
```
vinamrabenara, line 135: I was wondering if we should explicitly test here with ground truth rather than testing it against it's sibling

nums/core/linalg.py
```diff
@@ -256,6 +256,160 @@ def inv(app: ArrayApplication, X: BlockArray):
...
+
+    # Step 2: Compute U^-1
+    U_inv = app.inv_uppertri(U)
+
+    # Step 3: Compute inv(X) by U_inv @ U_inv.T
+    return U_inv @ U_inv.T
+
+
 def cholesky(app: ArrayApplication, X: BlockArray):
     # TODO (hme): Implement scalable version.
```
vinamrabenara, line 414: in my understanding, the ```cholesky_blocked``` is the scalable version for this? @elibol

nums/core/linalg.py
vinamrabenara: Apart from these, are there any assumptions you have made for the supported matrix types. Just want to be sure all the appropriate assertions are in place for what we support currently

nums/core/linalg.py
elibol: I think so, so long as the assumptions made about the input matrix to `choleskly_blocked` are the same as the NumPy implementation.

aatifjiwani: Is it okay if I keep it in this PR? May be a useful implementation down the line.

tests/core/test_linalg.py
aatifjiwani: @vinamrabenara Please tell me if you feel strongly about the explicit testing, but I believe the transitive law is at play here. In a previous test, we test that `linalg.inv == np.linalg.inv` with a tolerance of `1e-4`, so then intuitively it must mean that on line `135`, we are essentially using the explicit ground truth, but using a different API.

nums/core/linalg.py
aatifjiwani: Yes, besides the square matrix assumption, we are also making the assumption that the matrix is upper-triangular. I could implement validation for that, but would that be too much overhead? especially if `X` is super large?

tests/core/test_linalg.py
aatifjiwani: Sure!

nums/core/linalg.py
vinamrabenara: yes, looks good

tests/core/test_linalg.py
vinamrabenara: For now, just add a #TODO here for this

nums/core/linalg.py
vinamrabenara: hmm. what do you think @elibol

PR #155, opened 2021-09-25
Fix LinearRegression divergence bug

### LINKED ISSUE #141: LinearRegression model parameters diverge when fitting on integer BlockArrays
Likely because type of `_beta` is integer instead of float. Temporary fix is to cast inputs X and y to float.

Reproduce:
```python
import nums.numpy as nps
from nums.models.glms import LinearRegression
X = nps.array([[1, 2], [3, 4]])
y = nps.array([5, 6])
model = LinearRegression()
model.fit(X, y)
print(model._beta.get())  # array([-9223372036854775808, -9223372036854775808])
```

dlzou:
Closes #141

### DIFF
--- a/nums/models/glms.py
+++ b/nums/models/glms.py
@@ -111,6 +111,8 @@ class GLM(object):
         # by referencing X's existing blocks.
         # TODO: Option to do concat.
         # TODO: Provide support for batching.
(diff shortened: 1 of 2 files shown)


vinamrabenara (approved):
LGTM

PR #164, opened 2021-09-30
Add `tag` task

bveeramani:
This PR removes `version-tag.py` and replaces it with an [Invoke](http://www.pyinvoke.org/) task. You can invoke the task with `invoke tag` (or more concisely `inv tag`).


elibol (approved):
LGTM.

PR #165, opened 2021-09-30
Add Support for Scikit-Learn.

elibol:


elibol:
Update to lazy-initialize models.

nums/sklearn.py
```diff
@@ -0,0 +1,191 @@
...
+    from sklearn.preprocessing import StandardScaler, RobustScaler
+
+    sklearn_clses = [StandardScaler, RobustScaler]
+    return (build_sklearn_actor(sklearn_cls) for sklearn_cls in sklearn_clses)
+
+
+(StandardScaler, RobustScaler) = build_preprocessors()
+
+
+def exec_parallel(size, features):
```
dlzou, line 165: This looks pretty similar to code in test_sklearn.py, is it redundant?

nums/core/systems/systems.py
```diff
@@ -59,6 +63,16 @@ def register(self, name: str, func: callable, remote_params: dict = None):
     def call(self, name: str, args, kwargs, device_id: DeviceID, options: Dict):
         return self._remote_functions[name](*args, **kwargs)
 
+    def register_actor(self, name: str, cls: type):
+        assert name not in self._actors
```
vinamrabenara, line 67: shall we add a error message that the name is not unique?

nums/core/systems/systems.py
```diff
@@ -204,6 +220,24 @@ def num_cores_total(self) -> int:
...
+
+    def make_actor(self, name: str, *args, device_id: DeviceID = None, **kwargs):
+        # Distribute actors round-robin over devices.
+        if device_id is None:
+            device_id = self._devices[self._actor_node_index]
+            self._actor_node_index = (self._actor_node_index + 1) % len(self._devices)
+        actor = self._actors[name]
+        node = self._device_to_node[device_id]
+        node_key = self._node_key(node)
+        options = {"resources": {node_key: 1.0 / 10 ** 4}}
```
vinamrabenara, line 235: is 10**4 something specific or just a large number to accomodate many actors?

vinamrabenara (approved):

nums/core/systems/systems.py
elibol: Just a large number. I think working precision of resources is low (at the time I wrote this), which is why 10^4 is used and not something larger.

nums/core/systems/systems.py
elibol: This is low-level, we don't expect users to run this. If we make a mistake we want to know about it ;)

PR #170, opened 2021-10-09
Implementation of Upper-Triangular matrix inverse

aatifjiwani:
For our CS267 final project in Spring 2021, we implemented the algorithm from Bientenisi et al. in the NumS framework where we use Ray to take the inverse of an upper-triangular matrix.

This PR includes:

- An additional numpy API sum_reduce which is required for Upper-Triangular matrix inversion
- Implementation in linalg module for Upper-Tri matrix inversion

Compared to this [PR](https://github.com/nums-project/nums/pull/146), I have removed `cholesky_blocked`


tests/core/test_linalg.py
```diff
@@ -125,6 +125,19 @@ def test_inv(app_inst: ArrayApplication):
...
 
 
+def test_inv_uppertri(app_inst: ArrayApplication):
+    block_shape, shape = (3, 3), (20, 20)
+    R_np = np.arange(1, shape[0] ** 2 + 1, 1).reshape(*shape).astype(float)
+    R_np = np.triu(R_np)
+
+    R = app_inst.array(R_np, block_shape=block_shape)
+
+    R_inv_naive = np.linalg.inv(R_np)
```
aatifjiwani, line 135: @vinamrabenara explicit testing

aatifjiwani:
@vinamrabenara @elibol I would like to thank you for your patience with my PRs. Dealing with work and grad school applications hasn't been easy, so I appreciate your willingness to continue to review these PRs after so long.

nums/core/linalg.py
```diff
@@ -256,6 +256,147 @@ def inv(app: ArrayApplication, X: BlockArray):
...
+def inv_uppertri(app: ArrayApplication, X: BlockArray):
+    # Inversion of an Upper Triangular Matrix
+    # Use the method described in https://www.cs.utexas.edu/users/flame/pubs/siam_spd.pdf
+    assert len(X.shape) == 2
+    assert X.shape[0] == X.shape[1], "This function only accepts square matrices"
+    single_block = X.shape[0] == X.block_shape[0] and X.shape[1] == X.block_shape[1]
+    nonsquare_block = X.block_shape[0] != X.block_shape[1]
+
+    # If X is single block or block size is non-square, then reshape BS to row_size // 4
+    if single_block or nonsquare_block:
```
elibol, line 268: If X is a single block, why not return `inv`?

nums/core/linalg.py
```diff
@@ -256,6 +256,147 @@ def inv(app: ArrayApplication, X: BlockArray):
...
+        r11_oid, syskwargs={"grid_entry": (0, 0), "grid_shape": grid_shape}
+    )
+    R.blocks[(0, 0)].oid = r11_inv_oid
+    R_tl_shape = block_shape
+
+    # Continue while R_tl.shape != R.shape
+    while R_tl_shape[0] != full_shape[0] and R_tl_shape[1] != full_shape[1]:
+        # Calculate R11
+        R11_block = (
+            int(np.ceil(R_tl_shape[0] // block_shape[0])),
```
elibol, line 291: You're using integer division here, so I think the `int(np.ceil(...` portion is redundant.

nums/core/linalg.py
```diff
@@ -256,6 +256,147 @@ def inv(app: ArrayApplication, X: BlockArray):
...
+        for i, entry in enumerate(R01_grid_entries):
+            R.blocks[entry].oid = R01_2_oids[i]
+
+        # Recompute R_tl.shape
+        r11_r, r11_c = R11_shape
+        old_r, old_c = R_tl_shape
+
+        R_tl_shape = (old_r + r11_r, old_c + r11_c)
+
+    # By the time we finish, R = R_inv
```
elibol, line 396: Before returning, I think we'll need to reshape `R` to match the block shape of the input array `X`.

tests/core/test_linalg.py
```diff
@@ -125,6 +125,19 @@ def test_inv(app_inst: ArrayApplication):
         assert np.allclose(np.linalg.cholesky(mat.get()), L, rtol=1e-4, atol=1e-4)
 
 
+def test_inv_uppertri(app_inst: ArrayApplication):
```
elibol, line 128: Add assertions for tests here so that the input block shape matches the output block shape.

Also, can we include a test on less trivial input? I think you can obtain an invertible upper triangular matrix with `_, R = np.qr(np.random.rand(20, 20))`.

nums/core/linalg.py
elibol: Hmm, I see. Let's keep it the way it is for now. `inv` will be fast, but single-threaded. There might be a point beyond which what you're proposing will be faster, but we would need to test and apply heuristics accordingly (e.g. if the array is big then partition and use optimized imp).

elibol (commented):
Looking good! Could you update the code comments with standard English sentences with punctuation? Once the comments are finalized, I'll go ahead and merge!

Also, to follow up on discussion in the Cholesky PR: I agree checking that the matrix is upper-triangular is too much overhead. The checks we have are enough.

elibol (approved):
The partition diagram is super useful!

LGTM.

PR #178, opened 2021-10-17
Add Triu.

Priyansdesai:
I have added Triu. Test cases have been added.


nums/core/compute/numpy_compute.py
```diff
@@ -282,6 +282,25 @@ def advanced_assign_block_along_axis(
             dst_arr[tuple(dst_sel)] = src_arr[tuple(src_sel)]
         return dst_arr
 
+    def triu_copy(self, arr, rows, cols):
```
elibol, line 285: Can you explain what's going on here? is `arr.shape != (rows, cols)`? If it's equal, we can simply use `arr.copy()`, and in that case we ought to add a kernel op for copy.

elibol, line 285: Can you explain what's going on here? is `arr.shape != (rows, cols)`? If it's equal, we can simply use `arr.copy()`, and in that case we ought to add a kernel op for copy.

nums/core/array/application.py
```diff
@@ -439,6 +400,90 @@ def find_diag_output_blocks(X: BlockArray, total_elements: int):
             raise ValueError("X must have 1 or 2 axes.")
         return rarr
 
+    def _stack_copy(self, X):
```
elibol, line 403: Can you document this function (and all functions)? Include descriptions of what the function expects as input, what it outputs, and any additional details you think are worth mentioning.

nums/core/array/application.py
```diff
@@ -439,6 +400,90 @@ def find_diag_output_blocks(X: BlockArray, total_elements: int):
...
+                    X.blocks[block_row_index].oid, rows, cols, syskwargs=syskwargs
+                )
+                block_row_index += 1
+        return output_block_array
+
+    def triu(self, X: BlockArray):
+        if len(X.shape) == 1:
+            return self.triu(self._stack_copy(X))
+        elif len(X.shape) == 2:
+            if X.shape[0] == 1:
```
elibol, line 433: Add a comment explaining this case.

nums/core/compute/numpy_compute.py
```diff
@@ -282,6 +282,25 @@ def advanced_assign_block_along_axis(
...
+    def triu(self, arr, offset, zero_out, total_elements):
+        if not zero_out:
+            return np.triu(arr, k=offset)
+        else:
+            copy_arr = np.copy(arr)
+            if offset < 0:
+                copy_arr[:, :total_elements] = np.zeros((arr.shape[0], total_elements))
+            elif offset >= 0:
+                count = 0
+                while count < total_elements:
```
elibol, line 299: Is there a way to do this without entering a loop in Python?

nums/core/array/application.py
```diff
@@ -439,6 +400,90 @@ def find_diag_output_blocks(X: BlockArray, total_elements: int):
...
+                return X
+            diag_meta = array_utils.find_diag_output_blocks(X.blocks, min(X.shape))
+            output_arr_grid = ArrayGrid(X.shape, X.block_shape, X.dtype.__name__)
+            output_block_array = BlockArray(output_arr_grid, self.cm)
+            visited = dict()
+            total_row_blocks, total_col_blocks = X.blocks.shape[0], X.blocks.shape[1]
+            for block_indices, offset, total_elements in diag_meta:
```
elibol, line 440: Is there a single-pass solution, which does not require two passes over `diag_meta`?

nums/core/array/application.py
```diff
@@ -439,6 +400,90 @@ def find_diag_output_blocks(X: BlockArray, total_elements: int):
...
+                    X.blocks[block_indices].oid,
+                    offset,
+                    False,
+                    total_elements,
+                    syskwargs=syskwargs,
+                )
+                visited[block_indices] = 1
+            for block_indices, offset, total_elements in diag_meta:
+                row_c, col_c = block_indices[0] + 1, block_indices[1]
+                while row_c < total_row_blocks:
```
elibol, line 455: Could you explain what this inner-loop is doing?

nums/core/array/application.py
```diff
@@ -439,6 +400,90 @@ def find_diag_output_blocks(X: BlockArray, total_elements: int):
...
+    def triu(self, X: BlockArray):
+        if len(X.shape) == 1:
+            return self.triu(self._stack_copy(X))
+        elif len(X.shape) == 2:
+            if X.shape[0] == 1:
+                return X
+            diag_meta = array_utils.find_diag_output_blocks(X.blocks, min(X.shape))
+            output_arr_grid = ArrayGrid(X.shape, X.block_shape, X.dtype.__name__)
+            output_block_array = BlockArray(output_arr_grid, self.cm)
+            visited = dict()
```
elibol, line 438: Do we need to keep track of which blocks have been visited? Please explain why it's not possible to do without this dict.

elibol (commented):
Looking good! Left some comments.

Issue #152, opened 2021-09-17
Implement sort


elibol:
This is a common operation when working with climate data.

PR #258, opened 2022-01-29
Implement distributed sort

briancpark:
#152

Adapted from Stephanie Wang's [blog post](https://medium.com/distributed-computing-with-ray/executing-a-distributed-shuffle-without-a-mapreduce-system-d5856379426c). The code from the blog post is a version of [distributed bucket sort](https://en.wikipedia.org/wiki/Bucket_sort#:~:text=Bucket%20sort%2C%20or%20bin%20sort,applying%20the%20bucket%20sorting%20algorithm.&text=Sort%20each%20non%2Dempty%20bucket) using a map reduce design.

First, we sample pivots from each block within a BlockArray. These serves as pivots to sort into buckets. In the map phase, we sort each bucket, then split the array by the pivots to distribute into their respective buckets. In the reduce phase, we collect each sorted bucket outputted from the map phase and sort once again to give us a final sorted output. In the final step, we have to concatenate into a BlockArray.

We should be gaining speedup over NumPy since NumPy's implementation for sort is serial. This is by design as they only support multithreading for LAPACK/BLAS routines.


nums/core/array/application.py
```diff
@@ -1207,3 +1206,79 @@ def column_stack(self, tup, axis_block_size=None):
...
                 arrays.append(self.atleast_2d(arr))
         return self.concatenate(arrays, 1, axis_block_size=axis_block_size)
+
+    def sort(self, arr):
+        pivot_oids = arr.flattened_oids()
+        num_blocks = len(pivot_oids)
+
+        # If the BlockArray to be sorted is small enough, default to serial sort
+        if num_blocks == 1:
+            first_oid = self.cm.get(pivot_oids[0])
```
elibol, line 1216: No need to do `get` here, just pass in the oid!

nums/core/array/application.py
```diff
@@ -1207,3 +1206,79 @@ def column_stack(self, tup, axis_block_size=None):
             else:
                 arrays.append(self.atleast_2d(arr))
         return self.concatenate(arrays, 1, axis_block_size=axis_block_size)
+
+    def sort(self, arr):
+        pivot_oids = arr.flattened_oids()
```
elibol, line 1211: Would make it easier to follow code if this were named `block_oids`.

nums/core/array/application.py
```diff
@@ -1207,3 +1206,79 @@ def column_stack(self, tup, axis_block_size=None):
...
+                    "grid_shape": (1,),
+                    "options": {"num_returns": 1},
+                },
+            )
+            return BlockArray.from_oid(oid, arr.block_shape, arr.dtype, self.cm)
+
+        pivot_oids.pop()
+
+        # Sample the pivots to sort buckets around
+        pivots = []
```
elibol, line 1230: This ought to be `pivot_oids`.

nums/core/array/application.py
```diff
@@ -1207,3 +1206,79 @@ def column_stack(self, tup, axis_block_size=None):
...
+                arr_oid,
+                syskwargs={
+                    "grid_entry": (i,),
+                    "grid_shape": (num_blocks - 1,),
+                    "options": {"num_returns": 1},
+                },
+            )
+            pivots.append(BlockArray.from_oid(oid, (1,), arr.dtype, self.cm))
+
+        new_pivots = self.concatenate(pivots, 0, axis_block_size=num_blocks - 1)
```
elibol, line 1242: Would be easier to read if you used `axis_block_size=len(pivots)`.

nums/core/array/application.py
```diff
@@ -1207,3 +1206,79 @@ def column_stack(self, tup, axis_block_size=None):
...
+            mapped_sorted[i] = self.cm.map_sort(
+                arr_oids[i],
+                sorted_pivots_oid,
+                syskwargs={
+                    "grid_entry": (0,),
+                    "grid_shape": (1,),
+                    "options": {"num_returns": num_blocks},
+                },
+            )
+
```
elibol, line 1268: The block decomposition of the output here should be the same as the input. You can initialize an empty output BlockArray and assign the output oids to each block of the output block array. Might be worth double checking that the sort op is similar semantically to shuffle (for this to hold).

elibol (commented):
Looks great! Left some comments :)

nums/core/array/application.py
```diff
@@ -1206,3 +1205,89 @@ def column_stack(self, tup, axis_block_size=None):
...
+                    syskwargs={
+                        "grid_entry": grid_entry,
+                        "grid_shape": arr.grid.grid_shape,
+                        "options": {"num_returns": 1},
+                    },
+                )
+            )
+
+        for oid in oids:
+            shape = self.cm.get(oid).shape
```
elibol, line 1287: Here, you can call something like `shape_dtypes.append(self.cm.shape_dtype(oid, syskwargs={"grid_entry": grid_entry, "grid_shape": arr.grid.grid_shape}))` (use the same for loop as above). After the loop, call `self.cm.get(shape_dtypes)` to get a list of `(shape, dtype)` tuples. You can just use `zip` to transpose this into two lists of shapes and dtypes. You can then use the shapes in a final loop used to construct the individual block arrays.

PR #262, opened 2022-02-03
Refactor `DeviceID`

bveeramani:


elibol:
Relevant source: https://docs.python.org/3/library/dataclasses.html

elibol:
Did you figure out what's happening with this PR?

nums/core/grid/grid.py
```diff
@@ -105,64 +106,40 @@ def nbytes(self):
...
-
-    def __repr__(self):
-        return "%s=%s/%s:%s" % (
-            self.node_id,
-            self.node_addr,
-            self.device_type,
-            self.device_id,
-        )
-
-    def __eq__(self, other):
```
elibol, line 136: Can we define this in the dataclass? Should be defined as `self.node_id == other.node_id and self.device_id == other.device_id`. We do several thousand of these comparisons in the optimizer branch (which we will merge under experimental package soon) and I think a quick comparison of these integers will likely be faster than what dataclass may do by default (e.g. I expect perhaps it compares all properties).

If you think dataclass imp is faster, can you point to relevant documentation explaining how it compares equality by default?

elibol (commented):
Looks good for the most part. Left a minor comment.

elibol (approved):
LGTM

Issue #263, opened 2022-02-10
`test_modin.py` sometimes fails on workflow


briancpark:
I've seen this multiple times where `test_modin.py` will hang forever, specifically on Ray when running:
```sh
coverage run --parallel-mode -m pytest --system-name ray
```
Happens sporadically and I've seen it happen once per each version (3.7, 3.8, 3.9). Yet I cannot reproduce the error locally or know where the code is hanging. Could potentially be a deadlock depending on hardware resources.
Links:
* [1](https://github.com/nums-project/nums/runs/5133656237?check_suite_focus=true)
* [2](https://github.com/nums-project/nums/runs/5131032984?check_suite_focus=true)
* [3](https://github.com/nums-project/nums/runs/4975577050?check_suite_focus=true)
* [4](https://github.com/nums-project/nums/runs/4719840933?check_suite_focus=true)

PR #268, opened 2022-02-12
Add timeout to `test_modin`

bveeramani:
See #263

### DIFF
--- a/tests/test_modin.py
+++ b/tests/test_modin.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 import pytest
 
-
 from nums.core.array.blockarray import BlockArray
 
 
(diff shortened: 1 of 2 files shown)


tests/test_modin.py
```diff
@@ -11,12 +11,13 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import pytest
 
 from nums.core.array.blockarray import BlockArray
 
 
 # pylint: disable=import-outside-toplevel
+@pytest.mark.timeout(30)
```
briancpark, line 20: Will this be temporary once we find the bug? Seems like it's useful for debugging purposes but comes along with consequences. https://pypi.org/project/pytest-timeout/

If possible, you could try to rerun the workflows with a long for loop over `test_modin()`. Then see what the stack trace is if it happens to catch on a hang/timeout.

briancpark (commented):
Left a comment

bveeramani:
Closing in favor of #272

PR #270, opened 2022-02-12
Split `nums.core.backends.backends` into multiple modules

bveeramani:


elibol (commented):
This is great, but perhaps we ought to rename the modules to e.g. `raybackend` to avoid namespace collisions with the actual `ray` package.

bveeramani:
> This is great, but perhaps we ought to rename the modules to e.g. `raybackend` to avoid namespace collisions with the actual `ray` package.

I'm not sure if I entirely understand. Which collisions are you referring to? AFAIK there shouldn't be any issues with the `nums.core.backends.ray` module and `ray` package

elibol:
> > This is great, but perhaps we ought to rename the modules to e.g. `raybackend` to avoid namespace collisions with the actual `ray` package.
> 
> I'm not sure if I entirely understand. Which collisions are you referring to? AFAIK there shouldn't be any issues with the `nums.core.backends.ray` module and `ray` package

e.g. 

```
import ray
from nums.core.backends import ray
```

bveeramani:
> > > This is great, but perhaps we ought to rename the modules to e.g. `raybackend` to avoid namespace collisions with the actual `ray` package.
> > 
> > 
> > I'm not sure if I entirely understand. Which collisions are you referring to? AFAIK there shouldn't be any issues with the `nums.core.backends.ray` module and `ray` package
> 
> e.g.
> 
> ```
> import ray
> from nums.core.backends import ray
> ```

This shouldn't happen. At least the way I'm imagining it, the file structure of `nums.core.backends` is completely encapsulating. i.e., you should only call `from nums.core.backends import RayBackend`. Importing `nums.core.backends.ray` violates an abstraction layer

elibol (approved):
Reviewed again in light of discussion on namespace/variable binding collisions, and given the intended usage and structure of the backend package, I approve. We can always revisit if there's confusion /collisions.

Great idea! The code is much cleaner this way.

PR #284, opened 2022-02-21
Rename `Compute` to `Kernel`

bveeramani:


elibol (commented):
Looks great! Left a few comments. Might be worth doing a global (case-insensitive) search on `compute` to catch anything else you might have missed.

elibol (approved):
LGTM!

PR #292, opened 2022-02-25
Basic support for Zarr data.

elibol:


vinamrabenara:
no tests?

elibol:
> no tests?

Since this is an experimental API for read-only zarr file types, it is difficult to write a test without introducing zarr data structures into the repository. I think once we have more comprehensive support for the zarr type, we can add a read/write/delete test, and move it out of the experimental package.

vinamrabenara (approved):

PR #295, opened 2022-03-02
Add backend to nums.init.

elibol:


vinamrabenara (approved):
LGTM

nums/api.py
```diff
@@ -25,10 +25,14 @@ def init(
     address: Optional[str] = None,
     num_cpus: Optional[int] = None,
     cluster_shape: Optional[tuple] = None,
+    backend: Optional[str] = None,
 ):
     # pylint: disable = import-outside-toplevel
     import nums.core.settings as settings
 
+    if backend is not None:
+        assert backend in {"serial", "ray", "dask", "mpi"}
```
bveeramani, line 34: We should raise a `ValueError` instead of an `AssertionError`. From the Google style guide:
> Do not use assert statements for validating argument values of a public API. assert is used to ensure internal correctness, not to enforce correct usage nor to indicate that some unexpected event occurred.

(We use Black, but this is still applicable)

bveeramani: We should also include a descriptive error message if the users inputs an incorrect value. e.g., `f"Expected backend to be one of {BACKEND_NAMES}, but got {backend}."`

bveeramani (changes requested):
LGTM. Left a minor comment

PR #296, opened 2022-03-05
Fixes MPI tests to actually execute on more than one rank.

elibol:
@vinamrabenara tests configured with mpi backend were not actually configuring nums to use more than 1 rank, even though MPI may be launched with more than 1 rank.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ The following dependencies need to be installed (on Ubuntu or related Linux mach
 
 ```sh
 sudo apt update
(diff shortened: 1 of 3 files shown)


vinamrabenara:
> @vinamrabenara tests configured with mpi backend were not actually configuring nums to use more than 1 rank, even though MPI may be launched with more than 1 rank.

Oh, thanks for catching that @elibol !

vinamrabenara (approved):
LGTM

PR #299, opened 2022-03-10
Make `Kernel` an ABC

bveeramani:


bveeramani:
Closing because this isn't currently possible. We use magic to add methods, so a base class won't work here.

PR #302, opened 2022-03-14
Ga fusion

Priyansdesai:
Complete


nums/experimental/optimizer/fusion.py
```diff
@@ -67,7 +68,7 @@ def traverse(self, node: TreeNode, fuseable_nodes):
             # This branch is no longer fuseable.
             # We prune here and update fuseable_nodes with child nodes.
             fuseable_nodes += fuseable_children
-            return None, None
+            return None, None, 0
```
elibol, line 71: Is this a bug in the original code (returning 2 instead of 3 values)?

elibol (commented):
Looks good! Great job with this. Left a few comments. Do we need the changes you made to GraphArray? It looks functionally the same, so perhaps we can leave it as-is to minimize merge conflicts?

PR #343, opened 2022-04-05
Bump Black version to `22.3.0`

bveeramani:
Theses changes are needed to fix the Black CI:

image

The fix is implemented in the most recent version of Black. 

See https://github.com/psf/black/issues/2984.


briancpark (approved):
LGTM

PR #399, opened 2022-04-06
Alias `bitwise_not` with `invert`

bveeramani:
`bitwise_not` is not documented in the NumPy documentation (although, it is implemented).

For parity with NumPy, this PR:
1. Aliases `invert` with `bitwise_not`
2. Removes `bitwise_not` from the NumS documentation (`invert` remains in the NumS documentation).

### DIFF
--- a/docs/source/numpy.rst
+++ b/docs/source/numpy.rst
@@ -79,7 +79,6 @@ NumS Supported API
    arcsinh
    arctan
    arctanh
-   bitwise_not
    cbrt
    ceil
    conj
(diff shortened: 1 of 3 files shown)


docs/source/numpy.rst
```diff
@@ -159,4 +158,4 @@ NumS Supported API
    ones
    empty_like
    trace
-   top_k
\ No newline at end of file
+   top_k
```
bveeramani, line 161: EOL was missing here.

elibol (approved):
Looks like there's a minor conflict. Otherwise LGTM.

PR #436, opened 2022-04-06
Implement `asarray`

### LINKED ISSUE #308: Implement `asarray`

bveeramani:
Closes #308

### DIFF
--- a/nums/numpy/api/creation.py
+++ b/nums/numpy/api/creation.py
@@ -1116,3 +1116,77 @@ def zeros_like(prototype, dtype=None, order="K", shape=None):
     if order is not None and order != "K":
         raise NotImplementedError("Only order='K' is supported.")
     return zeros(shape, dtype)
+
+
+def asarray(a, dtype=None, order=None, *, like=None):
+    """Convert the input to an array.
+
+    Parameters
+    ----------
+    a : array_like
+        Input data, in any form that can be converted to an array.  This
+        includes lists, lists of tuples, tuples, tuples of tuples, tuples
+        of lists and ndarrays.
+    dtype : data-type, optional
+        By default, the data-type is inferred from the input data.
+
+    Returns
+    -------
+    out : ndarray
+        Array interpretation of `a`.  No copy is performed if the input
+        is already an ndarray with matching dtype and order.  If `a` is a
+        subclass of ndarray, a base class ndarray is returned.
+
+    Raises
+    ------
+    NotImplementedError
+        If you pass a non-`None` value to `order` or `like`.
+
+    See Also
+    --------
+    asanyarray : Similar function which passes through subclasses.
+    ascontiguousarray : Convert input to a contiguous array.
+    asfarray : Convert input to a floating point ndarray.
+    asfortranarray : Convert input to an ndarray with column-major
+                     memory order.
+    asarray_chkfinite : Similar function which checks input for NaNs and Infs.
+    fromiter : Create an array from an iterator.
+    fromfunction : Construct an array by executing a function on grid
+                   positions.
+
+    Examples
+    --------
+    Convert a list into an array:
+
+    >>> a = [1, 2]
+    >>> nps.asarray(a).get()
+    array([1, 2])
+
+    Existing arrays are not copied:
+
+    >>> a = nps.array([1, 2])
+    >>> nps.asarray(a) is a
+    True
+
+    If `dtype` is set, array is copied only if dtype does not match:
+
+    >>> a = nps.array([1, 2], dtype=np.float32)
+    >>> nps.asarray(a, dtype=np.float32) is a
+    True
+    >>> nps.asarray(a, dtype=np.float64) is a
+    False
+    """
+    if order is not None:
+        raise NotImplementedError("The `order` parameter isn't supported.")
+    if like is not None:
+        raise NotImplementedError("The `like` parameter isn't supported.")
+
+    if type(a) is BlockArray:  # pylint: disable=unidiomatic-typecheck
+        is_matching_dtype = not dtype or a.dtype == dtype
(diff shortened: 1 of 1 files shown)


nums/numpy/api/creation.py
```diff
@@ -1116,3 +1116,77 @@ def zeros_like(prototype, dtype=None, order="K", shape=None):
...
+    asfortranarray : Convert input to an ndarray with column-major
+                     memory order.
+    asarray_chkfinite : Similar function which checks input for NaNs and Infs.
+    fromiter : Create an array from an iterator.
+    fromfunction : Construct an array by executing a function on grid
+                   positions.
+
+    Examples
+    --------
+    Convert a list into an array:
```
briancpark, line 1156: Add

```
The doctests shown below are copied from NumPy.
They won’t show the correct result until you operate ``get()``.
```

nums/numpy/api/creation.py
```diff
@@ -1116,3 +1116,77 @@ def zeros_like(prototype, dtype=None, order="K", shape=None):
     if order is not None and order != "K":
         raise NotImplementedError("Only order='K' is supported.")
     return zeros(shape, dtype)
+
+
+def asarray(a, dtype=None, order=None, *, like=None):
+    """Convert the input to an array.
+
```
briancpark, line 1123: Add the lines:

```
This docstring was copied from numpy.asarray.

Some inconsistencies with the NumS version may exist.
```

This is just to be consistent with the other documentation.

nums/numpy/api/creation.py
```diff
@@ -1116,3 +1116,77 @@ def zeros_like(prototype, dtype=None, order="K", shape=None):
...
+        subclass of ndarray, a base class ndarray is returned.
+
+    Raises
+    ------
+    NotImplementedError
+        If you pass a non-`None` value to `order` or `like`.
+
+    See Also
+    --------
+    asanyarray : Similar function which passes through subclasses.
```
briancpark, line 1147: Delete other APIs we don't officially support.

briancpark (changes requested):
Requested some changes related to documentation.

nums/numpy/api/creation.py
bveeramani: I manually checked the docstring. There shouldn't be an inconsistencies

nums/numpy/api/creation.py
bveeramani: I deleted the two that we'll never support. The others we'll conceivably implement in the future. Do you think I should remove them now and add them back later, or just keep them?

nums/numpy/api/creation.py
bveeramani: I manually ran all the doctests. They all work correctly without changes

nums/numpy/api/creation.py
bveeramani: I figure if we delete them now we'll probably forget to add it back later

briancpark: If you want to implement them, then sure leave them behind. But I believe they'll link to NumPy docs, so make sure to recompile locally to make sure

nums/numpy/api/creation.py
briancpark: This is just a warning on all of the NumPy API docstrings

nums/numpy/api/creation.py
briancpark: This is just a warning on all of the NumPy API docstrings

nums/numpy/api/creation.py
```diff
@@ -1116,3 +1116,74 @@ def zeros_like(prototype, dtype=None, order="K", shape=None):
...
+    a : array_like
+        Input data, in any form that can be converted to an array.  This
+        includes lists, lists of tuples, tuples, tuples of tuples, tuples
+        of lists and ndarrays.
+    dtype : data-type, optional
+        By default, the data-type is inferred from the input data.
+
+    Returns
+    -------
+    out : ndarray
```
briancpark, line 1135: Change this to `BlockArray` and any other mentions of `ndarray` to `BlockArray`

nums/numpy/api/creation.py
bveeramani: You don't need to call `get` for two of the three examples, so I'm not sure if we should include it.

```
>>> a = nps.array([1, 2])
>>> nps.asarray(a) is a
True
```

```
 >>> a = nps.array([1, 2], dtype=np.float32)
 >>> nps.asarray(a, dtype=np.float32) is a
True
>>> nps.asarray(a, dtype=np.float64) is a
False
```

nums/numpy/api/creation.py
bveeramani: Okay, sounds good. I'll compile the documentation and check

nums/numpy/api/creation.py
bveeramani: Yeah. Was talking to @elibol about potentially renaming `BlockArray` to `ndarray`, but you're right. Shouldn't prematurely change the name.

bveeramani: Fixed

nums/numpy/api/creation.py
bveeramani: I hear you. I see how adding the lines improves consistency. That being said, I don't think we should warn users about inconsistencies if we know there are none.

briancpark: Sure, if @elibol approves, then I'm fine with not including this and resolving the followup with the doctest warning.

Other than that looks consistent with the documentation

elibol: sounds good to me.

elibol (approved):
LGTM.

PR #543, opened 2022-04-11
Change documentation theme to Executable Books

bveeramani:


nums/numpy/api/stats.py
```diff
@@ -808,7 +808,7 @@ def var(a: BlockArray, axis=None, dtype=None, out=None, ddof=0, keepdims=False):
     The doctests shown below are copied from NumPy.
     They won’t show the correct result until you operate ``get()``.
 
-    >>> a = nps.array([[1, 2], [3, 4]]) # doctest: +SKIP
+    >>> a = nps.array([[1, 2], [3, 4]]) # doctest: + SKIP
```
briancpark, line 811: I don't know if this matters, but looks like you pushed a minor spacing here.

docs/source/_toc.yml
```diff
@@ -0,0 +1,57 @@
...
+          - file: developer-guides/contributing-code/setting-up-your-development-environment
+          - file: developer-guides/contributing-code/testing-and-linting-your-code
+          - file: developer-guides/contributing-code/adding-documentation
+          - file: developer-guides/contributing-code/submitting-a-pull-request
+      - file: developer-guides/release-process
+
+  - caption: Community
+    chapters:
+      - url: https://google.com
+        title: Discord
```
briancpark, line 52: Is the plan to host a Discord? We already have a Slack that's meant for incoming users/collaboration, so maybe link the invitation link for that instead?

docs/source/index.md
```diff
@@ -0,0 +1,24 @@
...
+```
+
+## What documentation resource is right for me?
+
+```{include} _includes/_resources.md
+```
+
+## How can I get involved?
+
+NumS not only a [INSERT HERE], but also an active community of developers, researchers, and [INSERT HERE]
```
briancpark, line 17: Maybe try to resolve what to replace `INSERT HERE` with? It seems like an important part for visitors to get an impression of NumS :)

briancpark (changes requested):
Requested a few minor changes/suggestions. Other than that, I'm hoping you'll resolve the missing API documentation and missing links in follow up PRs. Looks really great so far!

nums/numpy/api/stats.py
bveeramani: Good catch. Fixed

docs/source/_toc.yml
bveeramani: I briefly talked to @elibol about this. I'm not sure if we should use Discord or Slack as the platform for our community. My current thinking is that we should we should use Discord because Discord is positioned as a platform for building communities, whereas Slack is positioned as a platform for doing work. Many open sources projects (e.g, Ethereum, OpenTK) use Discord. 

I don't think it's relevant that we already have a Slack since we have no users, and no one other than the development team uses the Slack. If we already had a community on Slack, that'd be different.

In any case, I think we can defer the decision and address this issue in a future PR. It's not like anyone's reading the documentation right now.

docs/source/index.md
bveeramani: I can't figure out what to write right now. I can fix it in soon in a future PR?

briancpark (approved):

PR #550, opened 2022-04-14
Document array manipulation routines

bveeramani:


docs/source/api-reference/array-manipulation-routines.rst
```diff
@@ -1,18 +1,14 @@
...
 
-.. currentmodule:: numpy
+.. currentmodule:: nums.numpy
 
 Basic operations
 ================
 .. autosummary::
    :toctree: generated/
 
+    copyto
```
briancpark, line 11: This isn't hyperlinked when I compile docs?

docs/source/api-reference/array-manipulation-routines.rst
```diff
@@ -28,6 +24,7 @@ Transpose-like operations
    :toctree: generated/
 
    swapaxes
+   ndarray.T
```
briancpark, line 27: `ndarray.T` doesn't show up at all when I compile locally.

docs/source/api-reference/array-manipulation-routines.rst
```diff
@@ -41,6 +38,13 @@ Changing number of dimensions
    expand_dims
    squeeze
 
+Changing kind of array
+======================
+.. autosummary::
+   :toctree: generated/
+
+   asarray
```
briancpark, line 46: `asarray` is not hyperlinked?

docs/source/api-reference/array-manipulation-routines.rst
```diff
@@ -60,14 +64,12 @@ Splitting arrays
...
-=============
+Adding and removing elements
+============================
 .. autosummary::
    :toctree: generated/
 
-   tile
-   repeat
-
+   append
```
briancpark, line 72: Same thing for append. Doesn't link...

briancpark (commented):
I pulled your fork, checked out out to `array-manipulation-routines` and compiled via `make html`. The hyperlinks don't show, and might be because we don't support them yet, although I thought we pushed `asarray` recently. I would advise against adding functions we don't support yet unless you have a good reason to or plan to implement them in the near future.

If I did something wrong in compiling process, lmk.

docs/source/api-reference/array-manipulation-routines.rst
bveeramani: It links if you merge #551

docs/source/api-reference/array-manipulation-routines.rst
bveeramani: See above

docs/source/api-reference/array-manipulation-routines.rst
bveeramani: See above

docs/source/api-reference/array-manipulation-routines.rst
briancpark: Gotcha, thanks for clarifying

briancpark (approved):
LGTM

docs/source/api-reference/array-manipulation-routines.rst
bveeramani: Good catch. I know `BlockArray.T` works, but I'm not sure where it's defined. @elibol do you know? Once I know where it's defined, I can add it to the docs

PR #564, opened 2022-09-07
add from_sparse method

trsigg:
Method for creating a `SparseBlockArray` from a `scipy.sparse` object.


dlzou:
@trsigg Can you remove the `asscalar` fallback so tests pass? Should be the same as #563. Also, can you add a test case for the new method?

dlzou (approved):

